PR #1, opened 2019-06-21
Fetch original image

SolitudeSF:
this fetches image in the max quality instead. which i think is the preferred behavior when clicking on the image.

### DIFF
--- a/src/views/tweet.nim
+++ b/src/views/tweet.nim
@@ -40,7 +40,7 @@
     #for photo in photos:
     <div class="attachment image">
       ##TODO: why doesn't this work?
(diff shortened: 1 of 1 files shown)


zestyr:
From my (limited) experience `:large` already achieves this, but it seems `?name=orig` works as well so this is fine for me. Do you have any documentation for this parameter?

SolitudeSF:
this is what youtube-dl/gallery-dl/whatever-userscript-for-fetching-images uses i think.

hmm, i checked few photos and `:large` indeed fetches the original, but that wasnt the case some time ago. i guess something changed internally in twitter.

zestyr:
so it turns out :large only fetches the original up until a certain size, where `?name=orig` is required. added to master, thanks

PR #8, opened 2019-08-08
Use local tmp directory and enhance readme

Perflyst:
Nitter should not use global tmp for more privacy and security. Also nitter should run under its own user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ WantedBy=multi-user.target
 ```
 
 Then enable and start
-`systemctl enable --now nitter.service && systemctl start nitter.service`
+`systemctl enable --now nitter.service`
 
 ## Todo (roughly in this order)
 
--- a/README.md
+++ b/README.md
@@ -14,14 +14,49 @@ Inspired by the [invidio.us](https://github.com/omarroth/invidious) project.
 ## Installation
 
 ```bash
-git clone https://github.com/zedeus/nitter
-cd nitter
-nimble build
+# useradd -m nitter
+# su nitter
+$ curl https://nim-lang.org/choosenim/init.sh -sSf | sh
+$ git clone https://github.com/zedeus/nitter
+$ cd nitter
+$ nimble build
+$ mkdir ./tmp
 ```
 
-To run, `./nitter`. It's currently not possible to change any settings or things
+To run nitter, execute `./nitter`. It's currently not possible to change any settings or things
 like the title, this will change as the project matures a bit. For now the focus
 is on implementing missing features.
+
+You should put nitter behind a reverse proxy with e.g. nginx or apache.
+
+It is also possibele to run nitter via systemd:
+
+```bash
+[Unit]
+Description=Nitter (An alternative Twitter front-end)
+After=syslog.target
+After=network.target
+
+[Service]
+Type=simple
+
+# set user and group
+User=nitter
+Group=nitter
+
+# configure location
+WorkingDirectory=/home/nitter/nitter
+ExecStart=/home/nitter/nitter/nitter
+
+Restart=always
+RestartSec=15
+
+[Install]
+WantedBy=multi-user.target
+```
+
+Than enable and start
+`systemctl enable nitter.service && systemctl start nitter.service`
 
 ## Todo (roughly in this order)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -14,15 +14,50 @@ Inspired by the [invidio.us](https://github.com/omarroth/invidious) project.
 ## Installation
 
 ```bash
-git clone https://github.com/zedeus/nitter
-cd nitter
-nimble build
+# useradd -m nitter
+# su nitter
+$ curl https://nim-lang.org/choosenim/init.sh -sSf | sh
```
zedeus: Instead of this, add a note above saying it requires a Nim installation with a link to the choosenim repo.

README.md
```diff
@@ -14,15 +14,50 @@ Inspired by the [invidio.us](https://github.com/omarroth/invidious) project.
...
+$ git clone https://github.com/zedeus/nitter
+$ cd nitter
+$ nimble build
```
zedeus: Add `-d:release` to the build command

README.md
```diff
@@ -14,15 +14,50 @@ Inspired by the [invidio.us](https://github.com/omarroth/invidious) project.
...
 ```
 
-To run, `./nitter`. It's currently not possible to change any settings or things
+To run nitter, execute `./nitter`. It's currently not possible to change any settings or things
 like the title, this will change as the project matures a bit. For now the focus
 is on implementing missing features.
 
+You should put nitter behind a reverse proxy with e.g. nginx or apache.
+
+It is also possibele to run nitter via systemd:
```
zedeus: Typo, "possible"

README.md
```diff
@@ -14,15 +14,50 @@ Inspired by the [invidio.us](https://github.com/omarroth/invidious) project.
...
+ExecStart=/home/nitter/nitter/nitter
+
+Restart=always
+RestartSec=15
+
+[Install]
+WantedBy=multi-user.target
+```
+
+Than enable and start
```
zedeus: Than -> Then

README.md
```diff
@@ -14,15 +14,50 @@ Inspired by the [invidio.us](https://github.com/omarroth/invidious) project.
...
+`systemctl enable nitter.service && systemctl start nitter.service`
```
zedeus: `systemctl enable --now nitter.service`

nitter.conf
```diff
@@ -5,5 +5,5 @@ title = "nitter"
 staticDir = "./public"
 
 [Cache]
-directory = "/tmp/niter"
-profileMinutes = 10  # how long to cache profiles
\ No newline at end of file
+directory = "./tmp"
```
zedeus: For security purposes, I think firejail is better suited for this. It's in `/tmp/` to keep everything in RAM, reducing disk load. I'm not an expert on this topic so please let me know if you think I'm wrong.

README.md
Perflyst: I chose https://nim-lang.org/install.html instead

nitter.conf
Perflyst: Firejail is not installed by default on all systems. Next to this `/tmp` is not RAM, it just will be flushed on a reboot. `/dev/shm` would be real RAM.

README.md
```diff
@@ -13,16 +13,53 @@ Inspired by the [invidio.us](https://github.com/omarroth/invidious) project.
 
 ## Installation
 
+You need to install nimble on your system: https://nim-lang.org/install.html
```
zedeus: Nim, not nimble. nimble is Nim's package manager which is used to build the project to pull in dependencies.

nitter.conf
zedeus: Fair enough, thank you.

README.md
```diff
@@ -55,8 +57,8 @@ RestartSec=15
 WantedBy=multi-user.target
 ```
 
-Than enable and start
-`systemctl enable nitter.service && systemctl start nitter.service`
+Then enable and start
+`systemctl enable --now nitter.service && systemctl start nitter.service`
```
zedeus: Adding `--now` starts the service as well, so you can remove the second part

Perflyst:
Please squash the pr if you merge it

PR #24, opened 2019-08-13
User prefs

### LINKED ISSUE #22: Start video only if focus (or play button)
At the moment all videos start playing automatically. This can slow down the browser tab if there a lot of videos.
I would like to have a play button or autoplay only on focus (with mouse pointer).

This could also be configureable, time for settings? ;)

zedeus:
Suggestions for preferences are welcome. Fixes #22

Todo:
- [x] More preferences
- [ ] Tests
- [x] CSS


Perflyst:
In case we can put suggestions here:
- Option to replace YouTube links with Invidious links (it should probably be invidio.us as long as there is no API for instances)

src/prefs.nim
```diff
@@ -0,0 +1,116 @@
...
+         label: "Enable hls.js video playback (requires JavaScript)",
+         defaultState: false),
+
+    Pref(kind: checkbox, name: "autoplayGifs", label: "Autoplay gifs",
+         defaultState: true),
+  ],
+
+  "Display": @[
+    Pref(kind: checkbox, name: "hideTweetStats",
+         label: "Hide tweet stats (replies, retweets, likes",
```
Perflyst:
```suggestion
         label: "Hide tweet stats (replies, retweets, likes)",
```

src/prefs.nim
```diff
@@ -0,0 +1,116 @@
...
+      options*: seq[string]
+    of input:
+      defaultInput*: string
+      placeholder*: string
+
+const prefList*: Table[string, seq[Pref]] = {
+  "Privacy": @[
+    Pref(kind: input, name: "replaceTwitter",
+         label: "Replace Twitter links with Nitter (blank to disable)",
+         defaultInput: "nitter.net", placeholder: "Nitter hostname"),
```
Perflyst: This should be the hostname of the nitter instance instead

src/nitter.nim
```diff
@@ -70,66 +80,84 @@ settings:
...
     if @"query".len == 0:
-      resp Http404, showError("Please enter a username.", cfg.title)
+      resp Http404, showError("Please enter a username.", cfg.title,
+                              getCookiePrefs(request))
     redirect("/" & @"query")
 
+  post "/saveprefs":
+    var prefs = getCookiePrefs(request)
+    genUpdatePrefs()
+    setCookie("preferences", $prefs.id, daysForward(360))
```
Perflyst: Cookie should be set as HttpOnly and Secure, more information: https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Secure_and_HttpOnly_cookies

Perflyst:
Suggestion: Add button to reset to default settings

src/prefs.nim
zedeus: It's a placeholder for the input field which is too small for that, I can change it to "Nitter instance"

Perflyst: I do not mean the placeholder. I mean the `defaultInput` should be the hostname of the nitter instance (e.g. if selfhosting nitter)

zedeus: Ah, I tried doing that but it's not really possible besides patching the code. The simple reason for this is that this (immutable) list of preferences is created at compile-time to generate a bunch of code, and the config is read at runtime so it's too late. I can provide a compile-time configuration, so you can set it like this `nimble build -d:hostname=nitter.net`

Perflyst: If this is not possible directly here, than what about adding a config file? A config file will probably be needed later anyways (for e.g. rss feeds?).

Perflyst: Would be an option, yes.

Perflyst:
Bug report:
- `www.youtube.com` is not covered while replacing with Invidious

Issue #30, opened 2019-08-28
Responsive design (mobile friendly)


Perflyst:
Currently nitter looks on mobile the same as on desktop and you need to zoom manually otherwise you cannot read anything.

Would it be possible to make nitter more responsive and mobile friendly?

PR #35, opened 2019-09-07
Add Viewport Meta in header

stom79:


Perflyst:
Would that fix #30?

stom79:
No. It needs also some works with media queries.

zedeus:
Thank you, this makes supporting mobile a lot easier, didn't know about this trick.

PR #39, opened 2019-09-13
Search

zedeus:


Perflyst:
`Load older` should not be shown if the result of this will be `No more items`

zedeus:
Unfortunately due to inconsistencies with the search endpoint that can't always be guaranteed

PR #73, opened 2019-10-09
Added a Dockerfile

pred2k:
Hi,
i've created a Dockerfile to avoid hassel with depencies to build nitter.

Tested with Docker for Mac 19.03.2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM nimlang/nim:1.0.0-ubuntu-onbuild as build
+FROM nimlang/nim:1.0.0-ubuntu as build
 
 ARG NITTER_HOSTNAME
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+ENV NITTER_HOSTNAME nitter.net
+
+WORKDIR /build
+
+COPY . /build
+
+RUN apt-get update && apt-get install -y libsass-dev
+
+RUN nimble build -d:release -d:hostname="$NITTER_HOSTNAME" --accept
```
pred2k: actually i'm not really sure if that replacement of `$NITTER_HOSTNAME` is working correctly

zedeus:
I appreciate your work, but @setenforce0 provided a cleaner solution along with CI. Prebuilt Docker images are now provided at https://hub.docker.com/r/zedeus/nitter
Thank you!

PR #94, opened 2019-11-18
Fix docker image

goodtiding5:
Use external volume to keep user data persistent.

Place 'nim' executable at '/usr/local/bin'.


entrypoint.sh
```diff
@@ -0,0 +1,35 @@
...
+set -e
+set -u
+
+BUILD=/build
+WORKD=/data
+
+build_working_dir() {
+    [ -f $WORKD/nitter.conf ] || cp -f  $BUILD/nitter.conf $WORKD/.
+    [ -d $WORKD/public ]      || cp -rf $BUILD/public      $WORKD/.
+    [ -d $WORKD/tmp ]         || mkdir -p $WOKRD/tmp
```
setenforce0: Typo
```suggestion
    [ -d $WORKD/tmp ]         || mkdir -p $WORKD/tmp
```

goodtiding5: Good catch!  I will make the fix and generate a new merge request.

zedeus: No need for a new request, just push the changes to your branch

goodtiding5: OK.  Done!

goodtiding5:
README file is updated.  Please review.

Perflyst:
Cache is stored in redis, the tmp folder is gone. If you want persistent configuration you can simply mount nitter.conf. I think this can be closed @zedeus

PR #166, opened 2020-05-25
Fix failing tests

setenforce0:
The chrome version in the travis containers was updated, and is incompatible with the chromedriver version. This PR manually installs the supported version of chrome and fixes the multiline tweet test, as the account of the test tweet went private.


zedeus:
Thanks!

Issue #157, opened 2020-05-09
Twitter is removing the Old UI, will this affect Nitter?


IamRifki:
Here's the notice that they gave.

![](https://user-images.githubusercontent.com/39834015/81482763-97895980-9263-11ea-97d7-0330cc358279.png)

zedeus:
That is bad news indeed. I have played around with the new public API that uses guest tokens just like the old video API, but I don't know how feasible it's going to be. Will definitely require a large rewrite of the nitter backend.

IamRifki:
I remembered that [twitter-scraper](https://github.com/bisguzar/twitter-scraper) supports the new UI, if I remember correctly.

IamRifki:
I'm going to guess this will massively setback all the other features until further notice, Why did you chose to scrape the old UI in the first place anyway?

zedeus:
I wasn't even aware the new UI existed when I started working on nitter.

IamRifki:
Actually nvm, they use the old UI.

zedeus:
For those who are wondering, I have been working on rewriting part of the backend to use the new JSON-based API, without requiring a developer account. It's _fast_, loading a full timeline page with profile and photo gallery in just ~400 ms vs Twitter loading the same thing in 1.3-1.5 seconds.

It's not done yet, but I'm making a lot of progress and it should be ready for deployment before June 1st. Here's my current todo list:
- [ ] Lists (easy)
- [ ] Preview media in quotes
- [ ] Promo video cards
- [ ] Parse profile bio mentions/hashtags
- [ ] Handle request error codes, show suspended
- [ ] Write proper Redis library and use it for caching
- [ ] Store tokens in Redis
- [ ] Improve general error handling code to prevent crashes
- [ ] Videos are apparently always available as mp4, new playback preferences?
- [ ] Bold search queries
- [ ] Consider using hover card API endpoint for profiles, it uses the "old" API with separate limits but same tokens, equally as fast if not faster, allows fetching multiple profiles at a time
- [ ] Move search bar below tabs

zedeus:
After a few hotfixes, https://nitter.net is now running Nitter v2, and it's going very well so far. Despite an incredible amount of traffic (10-30 tweets per _second_), it's as fast as a local instance, serving tweets in 2-300 ms and timelines in 3-600 ms. I did some fairly heavy stress tests on some endpoints a couple weeks ago so it should hold up, but I am a little worried.

roytam1:
[...]

PR #169, opened 2020-06-01
Nitter v2

zedeus:
Just in time to save us from #157, Nitter lives on to see another day.

Tons of changes, completely rewritten backend, many edgecases have been fixed and lots of effort has been put into making sure parsing and rendering is "correct". Redis is now used as a cache instead of SQLite, meaning you _must_ run a Redis instance for nitter. Host and port can be set in the config, along with some connection limits. There are some other new config options, but you likely don't need to change them, the defaults are fine.

Using Redis as a cache lets us cut out a lot of redundant requests, something Twitter's web client doesn't do. For example, fetching a timeline takes ~4-500 ms with nitter, while it takes upwards of 1.5 seconds with Twitter. Feel free to come talk to me on IRC about how the new backend works, if you have any questions, or general feedback less suited for a GitHub issue.

I apologize for not having this ready at least a few days ago for testing, but this took a ton of work since I started from scratch with reverse engineering the new undocumented API.


B0pol:
I can't build it, I've followed steps from readme
`nimble build -d:release` is fine (I had to update nim, but that's not the point here)
`nimble scss` is also fine
`mkdir ./tmp` same
`./nitter` give me this error:
```
asyncfutures.nim(383)    read
asyncfutures.nim(383)    read
asyncfutures.nim(383)    read
asyncfutures.nim(383)    read
asyncfutures.nim(383)    read
Error: unhandled exception: Software caused connection abort [OSError]
```
I guess it is related to Redis as you mentioned "you _must_ run a Redis instance for nitter". Therefore could you update the README steps?

zedeus:
Sure, haven't touched the readme yet

src/redis_cache.nim
```diff
@@ -17,8 +17,12 @@ proc setCacheTimes*(cfg: Config) =
...
 
 proc initRedisPool*(cfg: Config) =
-  pool = waitFor newRedisPool(cfg.redisConns, maxConns=cfg.redisMaxConns,
-                              host=cfg.redisHost, port=cfg.redisPort)
+  try:
+    pool = waitFor newRedisPool(cfg.redisConns, maxConns=cfg.redisMaxConns,
+                                host=cfg.redisHost, port=cfg.redisPort)
+  except OSError:
+    echo "Failed to connect to Redis."
+    quit()
```
B0pol, line 25: you should quit without a 0 return code if this error happened.

B0pol:
I've found a regression, see #171

B0pol:
I've found another regression, this time more important, see https://github.com/zedeus/nitter/issues/172

B0pol:
Another remark:
I've seen you have changed preview of retweets, and I really like it! It's way more visible, and we can even play the video / go to the image url without needing to click on the tweet.
Though I think there is a room for improvement:
image
The space between the the user name, display name, date and the video is really small, that would be great to enlarge it.

I've also noticed you have removed the duration: 
image
Is it a limitation from new api?

PR #184, opened 2020-06-08
Image URL encoding

somini:
The option is named `compatiblePicURL`, used base64.

I don't know enough Nim to get this to compile though. `@"encoded"` is not a string apparently, so the base64 decoder can't eat that.


zedeus:
Thank you! I appreciate the effort, but a proper solution required some refactoring and Jester trickery with regex. Hope 2542274 solves your problem :)

somini:
Thanks for this. I'll run my personal instance with this option turned on.

PR #251, opened 2020-09-03
systemd unit: use default.target instead of multi-user.target

ioistired:
Not all systems have their default.target aliased to multi-user.target.
End-user machines, for instance, may have it set to graphical.target.

systemd.special(7) defines default.target as
"the default unit systemd starts at bootup.
Usually, this should be aliased (symlinked) to
multi-user.target or graphical.target."

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ Restart=always
 RestartSec=15
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=default.target
 ```
 
 Then enable and run the service:


setenforce0:
Can you please expand on why this is an issue? From what I can tell, multi-user.target is always loaded before graphical.target, and it's not like you're going to want to run nitter in rescue mode.

Histalek:
> Can you please expand on why this is an issue? From what I can tell, multi-user.target is always loaded before graphical.target, and it's not like you're going to want to run nitter in rescue mode.

AFAIK this is not a real issue for system services, although it could be an option to respect the default target a given system is trying to reach. Regardless if it is a 'text-mode' oder 'gui-mode' system.

While it isn't an issue for a system service it would be an issue for a user service. Fedora 35 for example does not have a `multi-user.target` for user services:
```bash
[histalek@fedora35]$ systemctl --user cat multi-user.target
No files found for multi-user.target.
```

somini:
Sorry, this is wrong.

```
$ systemctl --user cat multi-user.target
No files found for multi-user.target.
$ systemctl status multi-user.target
● multi-user.target - Multi-User System
     Loaded: loaded (/usr/lib/systemd/system/multi-user.target; static)
     Active: active since Sun 2021-12-19 11:09:50 WET; 12h ago
       Docs: man:systemd.special(7)
```

The `multi-user` target doesn't have files, but it always exists.

The `default.target` is a symlink to the default target to hit. It doesn't make much sense to depend on it. `multi-user.target` is the correct target, since it's not something part of the boot process (in single-user mode), nor is it a graphical application.

Histalek:
> Sorry, this is wrong.
> 
> ```
> $ systemctl --user cat multi-user.target
> No files found for multi-user.target.
> $ systemctl status multi-user.target
> ● multi-user.target - Multi-User System
>      Loaded: loaded (/usr/lib/systemd/system/multi-user.target; static)
>      Active: active since Sun 2021-12-19 11:09:50 WET; 12h ago
>        Docs: man:systemd.special(7)
> ```
> The `multi-user` target doesn't have files, but it always exists.
You are missing the `--user` for the second command and are therefore talking to the system bus. Apparent by the path of the multi-user.target: `/usr/lib/systemd/system/multi-user.target`.

```bash
[histalek@fedora35]$ systemctl --user status multi-user.target
Unit multi-user.target could not be found.
```

> The `default.target` is a symlink to the default target to hit. It doesn't make much sense to depend on it. `multi-user.target` is the correct target, since it's not something part of the boot process (in single-user mode), nor is it a graphical application.

You are correct, in this case it doesn't need to be changed. I was only trying to argue that following the default target might still make sense, as it seems more universal than `multi-user.target` but to be honest i don't know if a distribution exists/could exist without the multi-user target.

somini:
@Histalek That's right. I would argue that nitter is always a system service. :grin: 

> i don't know if a distribution exists/could exist without the multi-user target.

`multi-user.target` is mentioned everywhere on the  SystemD docs, e.g:

https://www.freedesktop.org/software/systemd/man/systemd.service.html#

Unless distros randomly remove targets, it should work everywhere...

ioistired:
Alright.

PR #257, opened 2020-09-26
Add number of quoted replies

### LINKED ISSUE #250: Add number of quoted replies
Near number of comments, likes, retweets.

B0pol:
fixes #250


B0pol:
If anyone is looking, it should not be "play" but a "quote" icon and @zedeus told me he will add it

PR #258, opened 2020-10-01
Add a docker-compose file

FireMasterK:
This allows people to have a persistent Redis database when using docker. You will now need to run Redis outside of docker if you use docker, this is not required when running with docker-compose.

This also reduces the nitter image size from 40MB to just 10MB, mainly because the Redis image which is no longer included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,15 +100,23 @@ using the systemd service below. You should run Nitter behind a reverse proxy
 such as [Nginx](https://github.com/zedeus/nitter/wiki/Nginx) or Apache for
 security reasons.
 
+To run nitter with docker, you will need to install redis before you run these commands.
+
 To build and run Nitter in Docker:
 ```bash
 docker build -t nitter:latest .
-docker run -v $(pwd)/nitter.conf:/src/nitter.conf -d -p 8080:8080 nitter:latest
+docker run -v $(pwd)/nitter.conf:/src/nitter.conf -d --network host -p 8080:8080 nitter:latest
 ```
 
 A prebuilt Docker image is provided as well:
 ```bash
-docker run -v $(pwd)/nitter.conf:/src/nitter.conf -d -p 8080:8080 zedeus/nitter:latest
+docker run -v $(pwd)/nitter.conf:/src/nitter.conf -d --network host -p 8080:8080 zedeus/nitter:latest
+```
+
+Using docker-compose to run both nitter and redis as different containers:
+Change redisHost from `localhost` to `redis` in `nitter.conf` and then run:
+```bash
+docker-compose up -d
 ```
 
 Note the Docker commands expect a `nitter.conf` file in the directory you run them.
--- a/README.md
+++ b/README.md
@@ -100,20 +100,20 @@ using the systemd service below. You should run Nitter behind a reverse proxy
 such as [Nginx](https://github.com/zedeus/nitter/wiki/Nginx) or Apache for
 security reasons.
 
-To run nitter with docker, you will need to install redis before you run these commands.
+To run Nitter with docker, you will need to install Redis before you run these commands.
 
 To build and run Nitter in Docker:
 ```bash
 docker build -t nitter:latest .
-docker run -v $(pwd)/nitter.conf:/src/nitter.conf -d --network host -p 8080:8080 nitter:latest
+docker run -v $(pwd)/nitter.conf:/src/nitter.conf -d --network host nitter:latest
 ```
 
 A prebuilt Docker image is provided as well:
 ```bash
-docker run -v $(pwd)/nitter.conf:/src/nitter.conf -d --network host -p 8080:8080 zedeus/nitter:latest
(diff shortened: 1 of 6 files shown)


setenforce0 (approved):
Lgtm. Nice work 👍

FireMasterK:
I believe that this should also fix this error as the redis database wouldn't be restarted every time nitter crashes. (looks like a race condition)
```
redis.nim(126)           raiseRedisError
asyncfutures.nim(383)    read
asyncfutures.nim(383)    read
asyncfutures.nim(383)    read
redpool.nim(67)          migrateIter
asyncfutures.nim(383)    read
???                      initRedisPoolIter
asyncfutures.nim(383)    read
Error: unhandled exception: -LOADING Redis is loading the dataset in memory [RedisError]
```

@zedeus any chance of merging this?

danieljakots:
Thanks for working on that! It's cleaner to have redis in a separate container.
I'm using it with one improvement: I replaced FROM alpine:latest with alpine:3.12 as using the _latest_ tag is not deterministic. (The same could be said with FROM nimlang/nim:alpine as nim but I haven't got there yet).

FireMasterK:
I used the latest tag mainly for lesser maintenance work and also because none of these are containers are data-sensitive, if anyone has any objections, I'm happy to update the PR to use specific tags.

Perflyst:
I like this change but it will break existing deployments, especially because there are no versions and only the latest tage for the nitter image

FireMasterK:
True, but I still think it's worth merging this as it fixes this error: https://github.com/zedeus/nitter/pull/258#issuecomment-708901108

My current workaround is to recreate the container everytime it happens. (Every few weeks)

Only people with automatic updates will be affected.

README.md
```diff
@@ -94,15 +94,23 @@ using the systemd service below. You should run Nitter behind a reverse proxy
 such as [Nginx](https://github.com/zedeus/nitter/wiki/Nginx) or Apache for
 security reasons.
 
+To run nitter with docker, you will need to install redis before you run these commands.
```
zedeus, line 97: Nitter and Redis should be capitalized

nitter.conf
```diff
@@ -10,7 +10,7 @@ hostname = "nitter.net"
 directory = "./tmp"
 listMinutes = 240  # how long to cache list info (not the tweets, so keep it high)
 rssMinutes = 10  # how long to cache rss queries
-redisHost = "localhost"
+redisHost = "localhost" # Change to redis if using docker-compose
```
zedeus, line 13: Quotes for clarity: `Change to "redis" ...`

zedeus (changes requested):
@FireMasterK Sorry about the delay. Would you say this is ready to be merged after these changes, or is there something you would like to add/change?

FireMasterK:
I've made some minor changes along with that, it should still be good to merge :)

zedeus:
Excellent, thanks!

minus7:
> I like this change but it will break existing deployments, especially because there are no versions and only the latest tage for the nitter image

It did and it was *not* easy to find the cause. The readme does mention the external redis requirement, but I didn't think to look at that

zedeus:
@minus7 If you open an issue for this I can close and pin it so others with the same problem can see it quickly. Does that make sense? Sorry about the trouble.

FireMasterK:
@zedeus I created one at https://github.com/zedeus/nitter/issues/488

minus7:
Thanks!

zedeus:
Thank you! @FireMasterK

README.md
```diff
@@ -100,15 +100,23 @@ using the systemd service below. You should run Nitter behind a reverse proxy
 such as [Nginx](https://github.com/zedeus/nitter/wiki/Nginx) or Apache for
 security reasons.
 
+To run Nitter with docker, you will need to install Redis before you run these commands.
+
 To build and run Nitter in Docker:
 ```bash
 docker build -t nitter:latest .
-docker run -v $(pwd)/nitter.conf:/src/nitter.conf -d -p 8080:8080 nitter:latest
+docker run -v $(pwd)/nitter.conf:/src/nitter.conf -d --network host nitter:latest
```
selurvedu, line 108: > `--network host`

Using host network is a pretty bad advice. Those who choose to use containers should really know what they are doing, and if they don't, they should learn how to publish a port and how it's different from host networking rather than blindly copy-pasting a "presumably sane" command line from README.

There is no such problem in `docker-compose.yml`, it uses regular port publishing, just as README.md suggested before this change.

README.md
```diff
@@ -100,15 +100,23 @@ using the systemd service below. You should run Nitter behind a reverse proxy
...
 ```
 
 A prebuilt Docker image is provided as well:
 ```bash
-docker run -v $(pwd)/nitter.conf:/src/nitter.conf -d -p 8080:8080 zedeus/nitter:latest
+docker run -v $(pwd)/nitter.conf:/src/nitter.conf -d --network host zedeus/nitter:latest
```
selurvedu, line 113: Same here.

PR #265, opened 2020-10-19
[Dockerfile] Replace manual `apk update` w/ `--no-cache`

PeterDaveHello:
Just like the second apk command: `RUN apk --no-cache add pcre-dev sqlite-dev`!

Use `--no-cache` will save the `apk update` step, also leave no cache file and have a smaller image!


zedeus:
Thanks!

PR #276, opened 2020-11-10
Add a likes page

dundundunnn:
can you add a like page for nitter, also if ever possible a retweets page too


zedeus:
I think you mean to open an issue?

PR #332, opened 2021-01-30
Add automatic twitter theme

B0pol:
Based on browser them, it switches between light and dark.
Twitter theme has been renamed to twitter light.


Mennaruuk:
Awesomesauce!!

zedeus:
Please rename `twitter_light.css` back to `twitter.css` and add a new `auto_(Twitter).css` instead, that way it's slightly more intuitive imo and backwards compatible with currently selected themes

PR #353, opened 2021-03-09
Add Teddit redirects

### LINKED ISSUE #306: Replacement for Reddit: teddit.net and/or libredd.it
Hello,

Can you add [teddit.net](https://teddit.net) (old reddit ui) and/or [libredd.it](https://libredd.it/) (new reddit ui) for Reddit alternatives?

Thank you in advance!

FIGBERT:
Closes #306. Doesn't include Libreddit support, as i.redd.it and v.redd.it links include Teddit-specific logic.

### DIFF
--- a/nitter.conf
+++ b/nitter.conf
@@ -33,6 +33,7 @@ theme = "Nitter"
 replaceTwitter = "nitter.net"
 replaceYouTube = "invidio.us"
 replaceInstagram = ""
+replaceReddit = "teddit.net"
 proxyVideos = true
 hlsPlayback = false
 infiniteScroll = false
--- a/src/formatters.nim
+++ b/src/formatters.nim
@@ -6,6 +6,7 @@ const
   ytRegex = re"([A-z.]+\.)?youtu(be\.com|\.be)"
   twRegex = re"(www\.|mobile\.)?twitter\.com"
   igRegex = re"(www\.)?instagram.com"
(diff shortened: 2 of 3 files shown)


somini:
This doesn't work for libreddit? What's the missing URL?

FIGBERT:
It could be made to work with libreddit, but would require some changes. If you look at commit 98ce4ed6f0cdcc922d4f9d8c86317d3bff945a74, you see that the [i.redd.it]() and [v.redd.it]() redirects use Teddit-specific URL paths. libreddit uses different paths for video and images.

If we wanted to support both, we could do one of two things:
1. Not redirect [i.redd.it]() or [v.redd.it]() links.
2. Include a boolean setting that toggles between Teddit and libreddit and URL paths accordingly.

I went with Teddit because that's what I personally use, but I would be willing to do further work to support both projects.

somini:
I see, this requires more work. libreddit (or teddit) might accept a request to standardize on the image/video location.

https://github.com/spikecodes/libreddit/issues/147

It appears that Teddit has a more robust media implementation, so I think this MR should be good to merge.

zedeus:
Thanks @FIGBERT! I've decided to not support v.redd.it and i.redd.it for now, due to some issues described in the code's comments. I've also talked a bit about it here: https://github.com/spikecodes/libreddit/issues/371#issuecomment-1001274256

PR #406, opened 2021-07-03
Allow nitter to find .md and .svg files in non-default staticDir

duxovni:
Previously, even if `staticDir` was set to a value other than `./public`, Nitter still used hard-coded paths under `./public` for a few files.


src/views/about.nim
```diff
@@ -1,22 +1,21 @@
...
     h2: text "Instance info"
     p:
       text "Commit "
       a(href=link): text hash
 
-proc renderFeature*(): VNode =
+proc renderFeature*(cfg: Config): VNode =
   buildHtml(tdiv(class="overlay-panel")):
-    verbatim feature
+    verbatim markdown(readFile(cfg.staticDir / "md/feature.md"))
```
HeavenVolkoff, line 21: As far as my little understanding of Nim goes, putting the markdown and readFile calls inside the procedures would result in them running on every request for those routes, which is not the same behaviour as the previous code. Something along these lines below should work better:

```nim
var
  about: string
  feature: string
 
proc renderAbout*(cfg: Config): VNode =
  if (len(about) == 0):
    about = markdown(readFile(cfg.staticDir / "md" / "about.md"))
  buildHtml(tdiv(class="overlay-panel")):
    verbatim about
    h2: text "Instance info"
    p:
      text "Commit "
      a(href=link): text hash

proc renderFeature*(cfg: Config): VNode =
  if (len(feature) == 0):
    feature = markdown(readFile(cfg.staticDir / "md" / "feature.md"))
  buildHtml(tdiv(class="overlay-panel")):
    verbatim feature
```

duxovni: Oh, good catch, thank you!

aquarial:
This change would be very helpful for NixOS, because the `public` directory is placed in a store path like `/nix/store/d0pxxnqrlmixmdi64i15hvlx04df6s80-nitter-unstable-2021-07-18/share/nitter/public/`.

zedeus:
Thank you very much for the PR! Since I made some changes to how markdown files are used, I went with a different approach to solve the issue in a way that's consistent other parts of Nitter that use the config file

duxovni:
Awesome! This fixes the issue with about.md and deletes feature.md, but we're still looking for lp.svg in ./public , so I'll reopen this and remove the markdown-related changes.

zedeus:
lp.svg is loaded into the binary at compile-time where public should be in its normal place, so that shouldn't be an issue

duxovni:
Oh, huh, didn't realize `readFile` was evaluated at build time; thanks!

Issue #355, opened 2021-03-09
Send Content Security Policy (CSP) Header


pred2k:
Hi,
to avoid any accidental connection to twitter, i would suggest that nitter should send a strong CSP Header itself.

I use the following in my reverse proxy without breaking functionality in FireFox 86 and Chromium 89.

These are Preferences i've set in my `nitter.conf`:
```
proxyVideos = true
hlsPlayback = true
infiniteScroll = true
```

Content-Security-Policy:
```
default-src 'none';
img-src 'self';
style-src 'self' 'unsafe-inline';
script-src 'self' 'unsafe-inline' blob:;
font-src 'self';
media-src 'self' blob:;
base-uri 'none';
frame-ancestors 'none';
form-action 'self';
connect-src 'self';
upgrade-insecure-requests;
```

The `blob:` <scheme-source> are necessary for `hlsPlayback = true`.
`connect-src 'self';` was necessary for `infiniteScroll = true`.

zedeus:
There's no simple way to implement this without adding a lot of ugly code, and it's something reverse proxies can easily handle, as shown in the [example Nginx config](https://github.com/zedeus/nitter/wiki/Nginx).

PR #407, opened 2021-07-06
Add security headers and a strict Content Security Policy (CSP)

cyberhost-uk:
How the headers increase user security/privacy: 

- `x-frame-options` -  Indicates whether the browser should allow the webpage to be displayed in a frame within another webpage.
- `x-content-type-options` - Prevents Internet Explorer from MIME-sniffing a response away from the declared content-type.
- `referrer-policy` - Controls how much referrer information (sent via the Referrer header) should be included with requests. 
- `permissions-policy` - Controls what browser features can be used (The configuration added disables all features).
- `content-security-policy` (CSP) - Restricts where resources can be loaded from. This will prevent an accidental connection to twitter. 

More detail on the specific headers can be found at https://securityheaders.com

@pred2k Thanks for your example CSP provided with #355 

Related: #355

### DIFF
--- a/src/routes/media.nim
+++ b/src/routes/media.nim
@@ -54,7 +54,12 @@ proc proxyMedia*(req: jester.Request; url: string): Future[HttpCode] {.async.} =
       "Content-Type": res.headers["content-type", 0],
       "Content-Length": contentLength,
       "Cache-Control": maxAge,
-      "ETag": hashed
+      "ETag": hashed,
+      "x-frame-options": "DENY",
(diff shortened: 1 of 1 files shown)


DatAres37:
There's a missing space in `connect-src'self'` btw.

zedeus:
Thanks, but this doesn't actually help since it only applies to media (pictures and videos) rather than the main page request. I've decided these headers should be handled by the reverse proxy, not Nitter itself.

PR #409, opened 2021-07-21
RSS feeds for threads

### LINKED ISSUE #343: RSS feed for a thread?
For those threads with 300 tweets over 2 years.

duxovni:
Fixes #343


duxovni:
Previously, nitter used "does this page not have an RSS feed" as a proxy for "is this page a tweet", in order to decide what type of preview card to use. Now with my changes, user profiles will get the `summary_large_image` preview card type, causing the user's profile pic to be displayed as a large image in previews. I don't think that's especially a problem, but if people feel strongly that user profile preview cards should only have a thumbnail image, we can work out a better way of distinguishing them.

It'd also be neat to use the `player` card type for tweets with videos, like [robinuniverse/TwitFix](https://github.com/robinuniverse/twitfix) does, but extracting an actual mp4 URL from a tweet seems like a messy process that'd definitely want to be a separate PR.

PR #410, opened 2021-07-21
Make RSS feeds more readable

duxovni:
The tweets retrieved by Nitter use literal newline characters rather than HTML `<p>` or `<br>` tags. When these are included verbatim in RSS feeds, some RSS readers will just concatenate all the lines together, which can make tweets difficult to understand. The `pre-wrap` attribute used by the Nitter UI won't necessarily be respected by RSS readers, so this PR inserts `<br>` tags alongside newlines as a best-effort attempt to improve readability. In addition, it shortens the titles of RSS feed items to a truncated prefix of the tweet text, rather than having an awkwardly long title that duplicates the description. I can split those into two separate PRs if that's preferable.


Joshndroid:
I have been looking for a solution to the comically long titles in my rss reader for feeds, this could be it.

zedeus:
I don't use Nitter's RSS feature myself, so I have no strong opinions on this, but it looks good to me. Thank you!

PR #419, opened 2021-08-05
Adding Translation.

ManeraKai:
**Note:** This is an early version. I'll improve on it once it gets accepted.

The Arabic translation isn't that good. I just wrote it fast for testing purposes.
The settings page is built with macros. They're consts, I'll try to fix that problem.
I'll try to spend some time on the about page too.


IsmailHazem:
For anyone who wants to translate: this video will show how to access twitter with different languages without an account.
https://user-images.githubusercontent.com/40805353/128511691-fae455f6-0df7-4fee-a56f-461d3c2553a6.mp4

PR #429, opened 2021-08-19
Handle withheld tweets

HeavenVolkoff:
Currently nitter renders withheld tweets as if they were normal tweets, this is different from Twitter's behavior, as seen bellow:

image

image

This PR implements checks for the withheld properties sent by the Twitter API and changes the `.available` value according to them, so to correctly render these tweets. In addition I expanded the `.tombstone` logic to use the `.text` property whenever necessary, this is due to Twitter's API sending the tombstone text for withheld tweets in the `.text` property.

Also, I changed the render logic to also show quotes to unavailable tweets (If this is an undesired feature I can revert it).

This is the result:
image

The following spreadsheet was used to find accounts and tweets to test this change:
https://docs.google.com/spreadsheets/d/1ABJKoDKest3ZYny99_GrZ-nFsAPzNBpehfRAAYrjjpY/edit#gid=0


Perflyst:
Well, but I want to see those tweets too? Why should we censor nitter if the content is actually still available?

HeavenVolkoff:
> Well, but I want to see those tweets too? Why should we censor nitter if the content is actually still available?

The content is not actually available though. The Twitter API only sends the following generic message `@<username>'s account|tweet has been withheld in <location> in response|due to <reason>` int the `.text` property, and never sends the original tweet content as far as I tested. This PR is only focused in adjusting the styling of these tweets to match those of other unavailable tweets, no censoring is being implemented whatsoever. It is possible, in the future, to setup a system so that we can bypass tweets withheld in only some specific locations, but that would need some proxy shenanigans, that I think is out of the scope for this change.

src/parser.nim
```diff
@@ -268,6 +269,20 @@ proc parseTweet(js: JsonNode): Tweet =
         result.gif = some(parseGif(m))
       else: discard
 
+  let withheldInCountries = (
+    if js{"withheld_in_countries"}.kind == JArray:
+      js{"withheld_in_countries"}.to(seq[string])
+    else:
+      newSeq[string]()
+  )
```
zedeus, line 277: Add a blank line below this

nitter.nimble
```diff
@@ -17,7 +17,7 @@ requires "sass#e683aa1"
 requires "regex#2e32fdc"
 requires "nimcrypto >= 0.4.11"
 requires "markdown#abdbe5e"
-requires "packedjson#7198cc8"
+requires "packedjson#d11d167"
```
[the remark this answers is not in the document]
HeavenVolkoff: Yeah, I needed to update the package to have access to the implementation for the `.to()` method that was added in this new version, which I used here:
https://github.com/HeavenVolkoff/nitter/blob/master/src/parser.nim#L274

About any functionality removed from the new version of `packedjson`, I didn't saw anything relevant in the diff, as you can check here:
https://github.com/Araq/packedjson/compare/0.2.1...0.2.2

As far as I tested, I am running my private instance with this patch for a week and everything seemed to remain working fine.

src/parser.nim
HeavenVolkoff: done

zedeus:
Thanks!

PR #432, opened 2021-08-29
Remove hardcoded default Piped/Invidious hostname

acarasimon96:
Since e183a9c, I noticed that when I use `replaceYouTube=""` in `nitter.conf`, Nitter transforms YouTube links anyway, using piped.kavin.rocks for the domain name. I was expecting for YouTube links to be left alone if that setting is blank, so disabling the link transforms is impossible without cookies.

This PR (my very first one here) addresses this behavior and should not break YouTube link transforms when `replaceYouTube` is already filled in the config. This can also be useful if an instance chooses to disable transforming YouTube links in RSS feeds.

**Edit:** I mixed up the initial observations with how Nitter handles the blank `replaceYouTube` setting, and in fact both RSS feeds and the settings page did use the default domain when it is blank in the config file. I still believe that it's more appropriate to define the default value in `nitter.conf` rather than hardcoding it.

### DIFF
--- a/src/config.nim
+++ b/src/config.nim
@@ -33,9 +33,7 @@ proc getConfig*(path: string): (Config, parseCfg.Config) =
     redisPort: cfg.get("Cache", "redisPort", 6379),
     redisConns: cfg.get("Cache", "redisConnections", 20),
     redisMaxConns: cfg.get("Cache", "redisMaxConnections", 30),
-    redisPassword: cfg.get("Cache", "redisPassword", ""),
-
-    replaceYouTube: cfg.get("Preferences", "replaceYouTube", "piped.kavin.rocks")
+    redisPassword: cfg.get("Cache", "redisPassword", "")
   )
 
   return (conf, cfg)
--- a/src/prefs_impl.nim
+++ b/src/prefs_impl.nim
@@ -54,7 +54,7 @@ genPrefs:
       "Replace Twitter links with Nitter (blank to disable)"
       placeholder: "Nitter hostname"
 
-    replaceYouTube(input, "piped.kavin.rocks"):
+    replaceYouTube(input, ""):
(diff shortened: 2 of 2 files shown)


zedeus:
Hm this can't be right, it just removes the default. I'll have to check this out closer before merging. Thanks for the PR!

acarasimon96:
Thanks for looking at it. I made this PR because I was expecting Nitter to not convert YouTube links in RSS feeds when `replaceYouTube` is a blank string in the config file.

selurvedu:
I can confirm the issue.

acarasimon96:
@zedeus, it's been a month since you looked at this PR. Any update on this?

grmat:
> Hm this can't be right, it just removes the default.

Currently, the default piped.kavin.rocks entry is set in two places: the distributed nitter.conf and the prefs_impl.nim which makes it impossible to disable piped as an instance default. Leaving blank in nitter.conf will still fall back to piped.kavin.rocks. Deleting the fallback while keeping the default entry in the distributed nitter.conf will get consistent behaviour with the Bibliogram field and enable instances to disable YouTube redirection by default.

zedeus:
Bibliogram is empty by default because it was semi-broken at the time it was added, and now it's entirely broken. Piped works, and the goal of this project is to protect user privacy, so it will remain on by default. It's a bug that leaving it empty still uses the fallback, but the way you solved it is wrong.

acarasimon96:
> Bibliogram is empty by default because it was semi-broken at the time it was added, and now it's entirely broken. Piped works, and the goal of this project is to protect user privacy, so it will remain on by default.

You should've told me that in the first place, but either way, I just found a different approach to achieve that without removing the hardcoded defaults. I'm going to close this PR and carry on the conversation in a new one, which should appear below this comment once it's created.

PR #438, opened 2021-09-27
Add tweet card image to the timeline RSS feed

mooons:
Adding the card image (if any) to the timeline RSS feed, making the feed just a bit richer.

I've been using this on my own deployment for a good 10 days now. So far so good.

Please kindly review. :)

## Before

Without the patch, using https://nitter.net/WSJ/rss :

before

## After

With the patch, on my local deployment:

after

## Web

For comparison, this is what we have on the web UI, which does display the tweet card image:

web


zedeus:
Good job!

PR #439, opened 2021-09-29
Disable Piped by default

grmat:
Currently, the default piped.kavin.rocks entry is set in two places: the distributed nitter.conf and the prefs_impl.nim which makes it impossible to disable piped as an instance default. Leaving blank in nitter.conf will still fall back to piped.kavin.rocks. I deleted this fallback while keeping the default entry in the distributed nitter.conf to get consistent behaviour with the Bibliogram field.

### DIFF
--- a/src/prefs_impl.nim
+++ b/src/prefs_impl.nim
@@ -54,7 +54,7 @@ genPrefs:
       "Replace Twitter links with Nitter (blank to disable)"
       placeholder: "Nitter hostname"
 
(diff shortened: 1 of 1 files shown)


acarasimon96:
Hi there, thanks for your contribution. It appears that this same exact problem has already been covered by #432, which achieves the exact same result as this PR. I'm just waiting (impatiently) for the repo owner to get it merged...

grmat:
Whoops! Not trying to steal the show, I've just searched issues, not PRs. Thanks for the comment!

PR #445, opened 2021-10-02
Rework list api and use separator in rss redis key

### LINKED ISSUE #428: Instances sporadically get a random username's timeline for RSS feeds
In #424, there have been reports of some Nitter instances returning RSS feeds of a random username (typically Japanese accounts) instead of the requested user. Unfortunately, those comments are beyond the scope of the first comment, although they are totally related to the same user timeline RSS endpoint, so I've decided to continue the discussion here to keep clean the issue mentioned earlier.

## Example observed behavior

The following is based on [this recent comment by me](https://github.com/zedeus/nitter/issues/424#issuecomment-901357852).

Say I add the RSS feed for [@loveliveevents](https://nitter.kavin.rocks/loveliveeevents) (an account whose tweets are posted in English) to my feed reader app on my PC. (I use Akregator on Manjaro KDE, if you're curious.) A few days later, I would suddenly get a notification saying I got 16 new items because the instance returned the most recent tweets from [some random Japanese user](https://nitter.kavin.rocks/nakiriayame) through the same [feed URL](https://nitter.kavin.rocks/loveliveevents). A few hours later, the instance gets the correct username's tweets again.

## Further details and where we're at so far

This bug has affected some, if not all, Nitter instances, particularly the [nitter.pussthecat.org](https://nitter.pussthecat.org) public instance. I've discovered this same exact behavior at [nitter.kavin.rocks](https://nitter.kavin.rocks) not long after creating this issue, but it went back to normal after [the instance owner restored service due to the server running low on storage space](https://github.com/zedeus/nitter/issues/424#issuecomment-901380315).

@PeterDaveHello has recently [set up a new Tor-only instance](https://github.com/zedeus/nitter/issues/424#issuecomment-900416176) to investigate this bug, and I will be on the lookout when my private instance or any of the public instances I use does the same. In the meantime, I'll be learning some Nim so that I can contribute code to this project.

CC @FireMasterK @TheFrenchGhosty @zedeus

jackyzy823:
###  Rework on  list related api

`List's  name`( used for display)  cannot (and **shoud not**) be simply converted to `slug name` (used in URL) by replace " " with "-"

For example:

A display name `推特动物园` of `@starknight` cannot used as slug name ,it's slug name is  1434331244116938757 (same as list id)
another example
A display name `我的精神病家园` of `@cxiaoji` can be used as slugname. (maybe it is created in an early time)
A display name `VC A-List` 's slug name is "vc-a-list"

so we should use `/i/lists/<list_id>` as primary route and redirect `/<screen_name>/lists/<slug_name>`  to  `/i/lists/<list_id>`(the same as twitter do) 

Instead of ` /<screen_name>/lists/<display_name_or_some_slug_name_converted_from_display_name_by_adding_hyphen>/ `

#### UI improvement:
Set html title for `List` view.

###  use separator in rss redis key

for example :
previously a rss rediskey might be:  rss:abcdef123456  . 
So how to distinguish which part is username which part is cursor ?  It could be username: abcdef123 cursor:456 or username: abcdef cursor:123456

so we should add separator between this params. I choose `/` , but i think `:` is ok too. let me know what should be finally used.

~This **may** resolve #428 related issue.~

### DIFF
--- a/src/parser.nim
+++ b/src/parser.nim
@@ -73,7 +73,7 @@ proc parseGraphList*(js: JsonNode): List =
 
   result = List(
     id: list{"id_str"}.getStr,
-    name: list{"name"}.getStr.replace(' ', '-'),
+    name: list{"name"}.getStr,
     username: list{"user", "legacy", "screen_name"}.getStr,
     userId: list{"user", "legacy", "id_str"}.getStr,
     description: list{"description"}.getStr,
--- a/src/parser.nim
+++ b/src/parser.nim
@@ -75,7 +75,7 @@ proc parseGraphList*(js: JsonNode): List =
     id: list{"id_str"}.getStr,
     name: list{"name"}.getStr,
     username: list{"user", "legacy", "screen_name"}.getStr,
-    userId: list{"user", "legacy", "id_str"}.getStr,
+    userId: list{"user", "rest_id"}.getStr,
     description: list{"description"}.getStr,
     members: list{"member_count"}.getInt,
     banner: list{"custom_banner_media", "media_info", "url"}.getImageStr
--- a/src/redis_cache.nim
+++ b/src/redis_cache.nim
@@ -56,7 +56,7 @@ proc initRedisPool*(cfg: Config) {.async.} =
 
 template pidKey(name: string): string = "pid:" & $(hash(name) div 1_000_000)
 template profileKey(name: string): string = "p:" & name
-template listKey(l: List): string = toLower("l:" & l.username & '/' & l.name)
+template listKey(l: List): string = "l:" & l.id
 
 proc get(query: string): Future[string] {.async.} =
   pool.withAcquire(r):
@@ -129,17 +129,17 @@ proc getCachedPhotoRail*(name: string): Future[PhotoRail] {.async.} =
     result = await getPhotoRail(name)
     await cache(result, name)
 
-proc getCachedList*(username=""; name=""; id=""): Future[List] {.async.} =
-  let list = if id.len > 0: redisNil
-             else: await get(toLower("l:" & username & '/' & name))
+proc getCachedList*(username=""; slug=""; id=""): Future[List] {.async.} =
+  let list = if id.len == 0: redisNil
+             else: await get("l:" & id)
 
   if list != redisNil:
     result = fromFlatty(uncompress(list), List)
   else:
     if id.len > 0:
-      result = await getGraphListById(id)
+      result = await getGraphList(id)
     else:
-      result = await getGraphList(username, name)
+      result = await getGraphListBySlug(username, slug)
     await cache(result)
 
 proc getCachedRss*(key: string): Future[Rss] {.async.} =
--- a/src/routes/list.nim
+++ b/src/routes/list.nim
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: AGPL-3.0-only
-import strutils
+import strutils, uri
 
 import jester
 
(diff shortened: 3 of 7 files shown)


src/routes/list.nim
```diff
@@ -7,41 +8,43 @@ import ".."/[query, types, redis_cache, api]
...
+        title = "@" & list.username & "/" & list.name
         members = await getListMembers(list, getCursor())
-      respList(list, members, renderTimelineUsers(members, prefs, request.path))
-
-    get "/i/lists/@id/?":
-      cond '.' notin @"id"
-      let list = await getCachedList(id=(@"id"))
-      if list.id.len == 0:
-        resp Http404
-      await cache(list)
```
jackyzy823, line 46: Since cache is saved in previous getCachedList , it's not necessary to do it again here.

src/routes/list.nim
```diff
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-only
 import strutils
+import uri
```
zedeus, line 3:
```suggestion
import strutils, uri
```

src/parser.nim
```diff
@@ -61,13 +61,15 @@ proc parseGraphList*(js: JsonNode): List =
...
-    name: list{"name"}.getStr.replace(' ', '-'),
+    name: list{"name"}.getStr,
     username: list{"user", "legacy", "screen_name"}.getStr,
-    userId: list{"user", "legacy", "id_str"}.getStr,
+    userId: list{"user", "rest_id"}.getStr,
     description: list{"description"}.getStr,
     members: list{"member_count"}.getInt,
     banner: list{"custom_banner_media", "media_info", "url"}.getImageStr
   )
+  if result.banner.len == 0:
```
zedeus, line 71:
```suggestion

  if result.banner.len == 0:
```

src/routes/rss.nim
```diff
@@ -62,7 +62,7 @@ proc createRssRouter*(cfg: Config) =
 
       let
         cursor = getCursor()
-        key = $hash(genQueryUrl(query)) & cursor
+        key = "search/" & $hash(genQueryUrl(query)) & "/" & cursor
```
zedeus, line 65:
```suggestion
        key = "search:" & $hash(genQueryUrl(query)) & ":" & cursor
```

src/routes/rss.nim
```diff
@@ -82,7 +82,7 @@ proc createRssRouter*(cfg: Config) =
       let
         cursor = getCursor()
         name = @"name"
-        key = name & cursor
+        key = "twitter/" & name & "/" & cursor
```
zedeus, line 85:
```suggestion
        key = "tweets:" & name & ":" & cursor
```

src/routes/rss.nim
```diff
@@ -105,9 +105,9 @@ proc createRssRouter*(cfg: Config) =
         of "search": initQuery(params(request), name=name)
         else: Query(fromUser: @[name])
 
-      var key = @"name" & "/" & @"tab"
+      var key = "tab/" & @"name" & "/" & @"tab" & "/"
```
zedeus, line 108: For consistency, I think it makes sense to put the tab name in front here, so we have e.g. `rss:tweets:username`, `rss:media:username`, `rss:search:username`.

```suggestion
      var key = @"tab" & ":" & @"name" & ":"
```

src/routes/rss.nim
```diff
@@ -119,19 +119,18 @@ proc createRssRouter*(cfg: Config) =
...
       respRss(rss)
 
-    get "/@name/lists/@list/rss":
+    get "/i/lists/@id/rss":
       cond cfg.enableRss
-      cond '.' notin @"name"
       let
         cursor = getCursor()
-        key = @"name" & "/" & @"list" & cursor
+        key = "lists/" & @"id" & "/" & cursor
```
zedeus, line 126:
```suggestion
        key = 
          if cursor.len == 0: "lists:" & @"id"
          else: "lists:" & @"id" & ":" & cursor
```

src/routes/rss.nim
```diff
@@ -119,19 +119,18 @@ proc createRssRouter*(cfg: Config) =
...
 
       var rss = await getCachedRss(key)
       if rss.cursor.len > 0:
         respRss(rss)
 
       let
-        list = await getCachedList(@"name", @"list")
+        list = await getCachedList(id= (@"id"))
```
zedeus, line 133:
```suggestion
        list = await getCachedList(id=(@"id"))
```

src/routes/list.nim
```diff
@@ -8,41 +9,44 @@ import ".."/[types, redis_cache, api]
...
-    get "/@name/lists/@list/?":
+    get "/@name/lists/@slug/?":
       cond '.' notin @"name"
       cond @"name" != "i"
+      cond @"slug" != "memberships"
+      let
+        slug = decodeUrl(@"slug")
+        list = await getCachedList(@"name", slug)
+      if list.id.len == 0:
+        resp Http404
```
zedeus, line 32:
```suggestion
        resp Http404, showError("List \"" & @"slug" & "\" not found", cfg)
```

src/routes/list.nim
```diff
@@ -8,41 +9,44 @@ import ".."/[types, redis_cache, api]
 import ../views/[general, timeline, list]
 export getListTimeline, getGraphList
 
-template respList*(list, timeline, vnode: typed) =
+template respList*(list, timeline, title, vnode: typed) =
   if list.id.len == 0:
     resp Http404, showError("List \"" & @"list" & "\" not found", cfg)
```
zedeus, line 14:
```suggestion
    resp Http404, showError("List \"" & @"id" & "\" not found", cfg)
```

src/redis_cache.nim
```diff
@@ -119,17 +119,17 @@ proc getCachedPhotoRail*(name: string): Future[PhotoRail] {.async.} =
...
+             else: await get("l:" & id )
 
   if list != redisNil:
     result = fromFlatty(uncompress(list), List)
   else:
     if id.len > 0:
       result = await getGraphListById(id)
     else:
-      result = await getGraphList(username, name)
+      result = await getGraphList(username, slug)
```
zedeus, line 132: Since `slug` is now only for redirecting, these functions could be renamed to `getGraphList` and `getGraphListBySlug`

src/redis_cache.nim
```diff
@@ -119,17 +119,17 @@ proc getCachedPhotoRail*(name: string): Future[PhotoRail] {.async.} =
     result = await getPhotoRail(name)
     await cache(result, name)
 
-proc getCachedList*(username=""; name=""; id=""): Future[List] {.async.} =
-  let list = if id.len > 0: redisNil
-             else: await get(toLower("l:" & username & '/' & name))
+proc getCachedList*(username=""; slug=""; id=""): Future[List] {.async.} =
+  let list = if id.len == 0: redisNil
+             else: await get("l:" & id )
```
zedeus, line 124:
```suggestion
             else: await get("l:" & id)
```

src/parser.nim
```diff
@@ -61,13 +61,15 @@ proc parseGraphList*(js: JsonNode): List =
 
   result = List(
     id: list{"id_str"}.getStr,
-    name: list{"name"}.getStr.replace(' ', '-'),
+    name: list{"name"}.getStr,
     username: list{"user", "legacy", "screen_name"}.getStr,
-    userId: list{"user", "legacy", "id_str"}.getStr,
+    userId: list{"user", "rest_id"}.getStr,
```
zedeus, line 66: Are you sure about this one? I remember there being some weirdness going on with id_str vs rest_id

zedeus (changes requested):
Looks good, just some nits and RSS key changes

src/parser.nim
jackyzy823: According to my observation , there's no `id_str` under `legacy` now. so i had to choose `rest_id` to fill userId.

src/parser.nim
jackyzy823: Ok, removed.

zedeus:
Thanks!

PR #448, opened 2021-10-02
enableRSS configuration setting

### LINKED ISSUE #437: Disable RSS via configuration option
Can we get a configuration switch to disable RSS Feeds? I'm entirely willing to run a Nitter instance and keep it public but when I'm getting somewhere between 10-25k RSS requests in a single minute this endeavor becomes unfeasible, connections to redis happen so fast that when running redis-cli monitor I'm unable to distinguish individual words or patterns, it just flies by at incredible speed. The Bibliogram project which is basically Nitter for Instantgram has this feature and I think it is sorely needed. This would give the option for Nitter users to purposefully their instances for RSS requests or just Twitter proxying.

acarasimon96:
Resolves #437.

This new setting is self-explanatory. By default, `enableRSS` is set to true, which enables the RSS endpoint and shows a link to a timeline's RSS feed on the header. If set to false, Nitter will hide the RSS link from the header, and RSS endpoints should return a 404 Not Found error.

### DIFF
--- a/nitter.conf
+++ b/nitter.conf
@@ -28,6 +28,7 @@ tokenCount = 10
 # the limit gets reset every 15 minutes, and the pool is filled up so there's
 # always at least $tokenCount usable tokens. again, only increase this if
 # you receive major bursts all the time
+enableRSS = true  # set this to false to disable RSS feeds
 
 # Change default preferences here, see src/prefs_impl.nim for a complete list
 [Preferences]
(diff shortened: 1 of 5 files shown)


Salastil:
Doesn't seem to work if you manually type in domain.tld/username/rss which would leave this route active.

zedeus:
Thanks!

Issue #295, opened 2020-12-07
Support /i/user/[userid] urls


0x7D2B:
Twitter has a URL scheme which allows linking to users using their id, rather than username: ` /i/user/[userid]`

For example, https://twitter.com/i/user/20603690 will redirect to https://twitter.com/korindo. It would be quite nice to have Nitter support this as well.

Looking at Twitter itself, the web app will GET https://api.twitter.com/1.1/users/show.json?include_profile_interstitial_type=1&include_blocking=1&include_blocked_by=1&include_followed_by=1&include_want_retweets=1&include_mute_edge=1&include_can_dm=1&include_can_media_tag=1&skip_status=1&user_id=20603690. The response will include `screen_name`, which will be the username of the user. Nitter actually already uses `show.json` in `getProfile`, but with the username (screen_name) as the argument.

zedeus:
Solved by #450

PR #450, opened 2021-10-04
Implement user_id to screen_name router

jackyzy823:
Make a  redirection of  `/i/user/<user_id>` or `/intent/user?user_id=<user_id>` to  screen_name link `/<screen_name>

Add a new redis key for mapping from `user_id` to  `screen_name`

When fetching profile via user_id , we also cache the profile.
When caching the profile (via user_id or screen_name)  , we also cache the mapping from `user_id` to `screen_name`.

for #295

### DIFF
--- a/src/api.nim
+++ b/src/api.nim
@@ -40,6 +40,12 @@ proc getProfile*(username: string): Future[Profile] {.async.} =
     url = userShow ? ps
   result = parseUserShow(await fetch(url, oldApi=true), username)
 
+proc getProfileById*(userId: string): Future[Profile] {.async.} =
+  let
+    ps = genParams({"user_id": userId})
+    url = userShow ? ps
+  result = parseUserShowId(await fetch(url, oldApi=true), userId)
+
 proc getTimeline*(id: string; after=""; replies=false): Future[Timeline] {.async.} =
   let
     ps = genParams({"userId": id, "include_tweet_replies": $replies}, after)
--- a/src/redis_cache.nim
+++ b/src/redis_cache.nim
@@ -78,6 +78,7 @@ proc cache*(data: Profile) {.async.} =
   pool.withAcquire(r):
     r.startPipelining()
     discard await r.setex(name.profileKey, baseCacheTime, compress(toFlatty(data)))
+    discard await r.setex("i:" & data.id , baseCacheTime, data.username)
     discard await r.hset(name.pidKey, name, data.id)
     discard await r.flushPipeline()
 
@@ -109,6 +110,15 @@ proc getCachedProfile*(username: string; fetch=true): Future[Profile] {.async.} 
     result = fromFlatty(uncompress(prof), Profile)
   elif fetch:
     result = await getProfile(username)
+
+proc getCachedProfileScreenName*(userId: string): Future[string] {.async.} =
+  let username = await get("i:" & userId)
+  if username != redisNil:
+      result = username
+  else:
+    let profile = await getProfileById(userId)
+    result = profile.username
+    await cache(profile)
 
 proc getCachedPhotoRail*(name: string): Future[PhotoRail] {.async.} =
   if name.len == 0: return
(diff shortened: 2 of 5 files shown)


src/routes/timeline.nim
```diff
@@ -105,8 +105,21 @@ template respTimeline*(timeline: typed) =
     resp Http404, showError("User \"" & @"name" & "\" not found", cfg)
   resp t
 
+template respUserid*() = 
+  cond @"user_id".len > 0
+  let screen_name = await getCachedProfileScreenName(@"user_id")
+  if screen_name.len > 0:
+    redirect("/" & screen_name)
+  else:
+    resp Http404
```
zedeus, line 114:
```suggestion
    resp Http404, showError("User not found", cfg)
```

src/routes/timeline.nim
```diff
@@ -105,8 +105,21 @@ template respTimeline*(timeline: typed) =
...
+
 proc createTimelineRouter*(cfg: Config) =
   router timeline:
+    get "/i/user/@user_id":
```
zedeus, line 118:
```suggestion
    get "/i/user/@id":
```

src/routes/timeline.nim
```diff
@@ -105,8 +105,21 @@ template respTimeline*(timeline: typed) =
...
+      respUserid()
+    get "/intent/user":
```
zedeus, line 120: Is this route missing a parameter? Also needs an empty line above it

src/routes/unsupported.nim
```diff
@@ -11,10 +11,12 @@ proc createUnsupportedRouter*(cfg: Config) =
       resp renderMain(renderFeature(), request, cfg, themePrefs())
 
     get "/about/feature": feature()
-    get "/intent/?@i?": feature()
+    get "/intent/?@i?":
```
zedeus, line 14: Now that this has a body with more than one line it should go below the `@name/lists` route with empty lines around it

src/api.nim
```diff
@@ -40,6 +40,12 @@ proc getProfile*(username: string): Future[Profile] {.async.} =
     url = userShow ? ps
   result = parseUserShow(await fetch(url, oldApi=true), username)
 
+proc getProfileById*(userid: string) :Future[Profile] {.async.} =
```
zedeus, line 43:
```suggestion
proc getProfileById*(userid: string): Future[Profile] {.async.} =
```

zedeus (changes requested):
All instances of `screen_name` and similar should be `username` for consistency. `userid`/`Userid` -> `userId`/`UserId`

src/routes/timeline.nim
jackyzy823: I'm not very familiar with Jester.

I think  "/i/user/@user_id"  and  "/intent/user"?user_id=xxx will both set @"user_id" parameter.

zedeus: Yeah I think that's correct, it's a little weird but Jester does indeed merge the URL parameters and query parameters.

src/redis_cache.nim
```diff
@@ -110,6 +111,15 @@ proc getCachedProfile*(username: string; fetch=true): Future[Profile] {.async.}
   elif fetch:
     result = await getProfile(username)
 
+proc getCachedProfileScreenName*(userId: string): Future[string] {.async.} =
```
zedeus, line 114:
```suggestion
proc getCachedProfileUsername*(userId: string): Future[string] {.async.} =
```

src/redis_cache.nim
```diff
@@ -110,6 +111,15 @@ proc getCachedProfile*(username: string; fetch=true): Future[Profile] {.async.}
...
+  let username = await get("i:" & userId)
+  if username != redisNil:
+      result = username
```
zedeus, line 117:
```suggestion
    result = username
```

PR #467, opened 2021-11-20
Fix #466 for a new cardkind newsletter_publication

### LINKED ISSUE #466: Unable to view a Profile in any Instance
nitter.net/user/startpage - Official Twitter Account of StartPage Search. It is not loading in any of the instances instead showing an error like this. Why is this happening ?

Error Message :

```
An error has occured in one of your routes.

Detail: strutils.nim(1315) parseEnum
asyncfutures.nim(372) read
asyncfutures.nim(372) read
asyncfutures.nim(372) read
asyncfutures.nim(372) read
asyncfutures.nim(372) read
Invalid enum value: newsletter_publication
```

jackyzy823:
Just a quick fix for #466 

Maybe we should have a default fallback for all `parseEnum`.

### DIFF
--- a/src/parser.nim
+++ b/src/parser.nim
@@ -169,7 +169,7 @@ proc parseCard(js: JsonNode; urls: JsonNode): Card =
   let
     vals = ? js{"binding_values"}
     name = js{"name"}.getStr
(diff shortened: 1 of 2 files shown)


zedeus:
Thanks!

PR #494, opened 2021-12-30
fix datetime fromFlatty

jackyzy823:
Ref: https://github.com/treeform/flatty/blob/master/tests/test_datetime.nim

This cause stuck when deserializing `joinDate` of `Profile` in `Recommendations` which i'm working on.


zedeus:
Thanks!

PR #503, opened 2022-01-05
Tweet stats tweak

LainLayer:
- adjusted tweet-stat css to improve looks
- removed the number from stats when its 0


zedeus:
Thank you, please bump the style version in general.nim

Issue #479, opened 2021-12-17
square avatars


ckp95:
can you make avatars be square like they used to be? the circles are when twitter started to go downhill.

the avatar on someone's profile is already square but the tweet ones are circles which is kind of inconsistent.

![image](https://user-images.githubusercontent.com/43073868/146564371-09e64070-c5ac-4120-bd96-e739c341c695.png)

zedeus:
I'd happily accept a PR with a settings toggle for square avatars.

LainLayer:
done :)

Mennaruuk:
@LainLayer Avatars are still round :( Only ones in search results show up as square.

![B8AF4BF1-4A10-45F0-8006-D7A8B5871EC9](https://user-images.githubusercontent.com/52135169/149438931-30ac6589-412d-4242-8ce0-4441aca2da12.jpeg)

PR #513, opened 2022-01-13
Implemented square profile pictures preference

LainLayer:
Added preference "Square profile pictures"

Implements #479


zedeus:
Thanks!

PR #527, opened 2022-01-23
Enhance the compose file

TheFrenchGhosty:
- Add persistence to redis, following the official documentation ( https://hub.docker.com/_/redis see "start with persistent storage")
- Bind to localhost by default, so that the port isn't unnecessarily exposed
- Make nitter `depends-on` its redis container
- Rename the redis container so that it's nitter specific (edit the example config file accordingly)
- Enforce `container_name` (if you don't want that, feel free to remove those lines)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
     ports:
       - "127.0.0.1:8080:8080" # Replace with "8080:8080" if you don't use a reverse proxy
     volumes:
-      - ./nitter.conf:/src/nitter.conf
+      - ./nitter.conf:/src/nitter.conf:ro
     depends_on:
       - nitter-redis
     restart: unless-stopped
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,18 +1,25 @@
-version: "3.8"
+version: "3"
+
 services:
-  redis:
-    image: redis:6-alpine
-    restart: unless-stopped
-    volumes:
-      - redis-data:/var/lib/redis
+
   nitter:
     image: zedeus/nitter:latest
-    restart: unless-stopped
-    depends_on:
-      - redis
+    container_name: nitter
     ports:
-      - "8080:8080"
+      - "127.0.0.1:8080:8080" # Replace with "8080:8080" if you don't use a reverse proxy
     volumes:
       - ./nitter.conf:/src/nitter.conf
+    depends_on:
+      - nitter-redis
+    restart: unless-stopped
+
+  nitter-redis:
+    image: redis:6-alpine
+    container_name: nitter-redis
+    command: redis-server --save 60 1 --loglevel warning
+    volumes:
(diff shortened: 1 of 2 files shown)


zedeus:
@FireMasterK please review

docker-compose.yml
```diff
@@ -1,18 +1,25 @@
-version: "3.8"
+version: "3"
```
FireMasterK, line 1: The version field is deprecated as per the compose specification now: https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-top-level-element, maybe we could remove this now?

docker-compose.yml
```diff
@@ -1,18 +1,25 @@
...
     ports:
-      - "8080:8080"
+      - "127.0.0.1:8080:8080" # Replace with "8080:8080" if you don't use a reverse proxy
     volumes:
       - ./nitter.conf:/src/nitter.conf
+    depends_on:
+      - nitter-redis
+    restart: unless-stopped
+
+  nitter-redis:
```
FireMasterK, line 16: Why change the name of the service? Wouldn't this break existing installs if they decided to `git pull`?

FireMasterK (approved):
Looks good! Just a few minor nit-picks.

docker-compose.yml
TheFrenchGhosty: I'm aware that it's deprecated, but older version of compose still need it, so for now, it's a good idea to keep it.

docker-compose.yml
TheFrenchGhosty: >Rename the redis container so that it's nitter specific

It make it so that another software won't use it.

It won't be a problem, since currently it did "nothing" since it was all in RAM.

docker-compose.yml
FireMasterK: Looks like you're right, version 1.28 works fine, however, 1.26 seems to choke saying "Unsupported config option for ..."

ACodingGenie:
If you guys want, I have a few options which might be worth adding to the nitter container. I've personally been using these for a while, and they work fantastic. Not sure if the cap_drop is necessary with user:nobody but idk...
```
    cap_drop:
      - ALL
    user: nobody
    read_only: true
     volumes:
      - ./nitter.conf:/src/nitter.conf:ro # Mount nitter.conf as read only
```

TheFrenchGhosty:
@ACodingGenie 

I don't think `cap_drop` is needed since I don't think that containers don't have any "cap" unless you `cap_add` them.

---

`user: nobody` is basically useless, since the container runs as root anyway.

---

Having the configuration file as read only, isn't useful... but I guess why not.

---

According to the first information I found about it ( https://docs.solace.com/Security/Security-Solace.htm ) `security-opt="no-new-privileges:true"` "prevent the container processes from gaining new privileges" I don't see how this is useful... since again, it runs as root, and it's not exposed anyway.

PR #529, opened 2022-01-24
Fix bio encoding issue

### LINKED ISSUE #521: Encoding problems in bio
Bios which contain (certain?) non-ASCII characters (only "ü"?) get scrambled.

Example: https://nitter.net/lu_eder -- "fünffachen" gets replaced by `f�<a href="/search?q=%23nffachen">�nffachen</a>` (note also the generated link that is not existing in the original bio).

Other example: https://nitter.net/PatzigProf -- "darüber" gets replaced by `dar�<a href="/search?q=%23ber">�ber</a>`

LainLayer:
image

My theory is `＃` inside `[]` is treated like a set of characters as its actually 3 characters long being UTF-8
instead, its compared like a string.

Fixes #521

### DIFF
--- a/src/parserutils.nim
+++ b/src/parserutils.nim
@@ -210,7 +210,7 @@ proc expandUserEntities*(user: var User; js: JsonNode) =
 
   replacements.deduplicate
   replacements.sort(cmp)
-  
+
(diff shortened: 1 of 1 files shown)


zedeus:
Thanks!

PR #531, opened 2022-01-27
Utilize docker cache for less downloads

gardenrobot:
Hi. This PR takes advantage of the docker cache. Now, a system will not have to redownload the dependencies on every build.

I'm not very familiar with nimble build commands, so any comments or requested changes are very welcome.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,10 +4,13 @@ EXPOSE 8080
 
 RUN apk --no-cache add libsass-dev pcre
 
-COPY . /src/nitter
 WORKDIR /src/nitter
 
-RUN nimble build -y -d:danger -d:lto -d:strip \
+COPY nitter.nimble .
+RUN nimble install -y -d:strip
+
+COPY . .
+RUN nimble build -y -d:danger -d:lto \
     && nimble scss \
     && nimble md
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ COPY nitter.nimble .
 RUN nimble install -y -d:strip
 
 COPY . .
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -4,10 +4,13 @@ EXPOSE 8080
...
 
-COPY . /src/nitter
 WORKDIR /src/nitter
 
-RUN nimble build -y -d:danger -d:lto -d:strip \
+COPY nitter.nimble .
+RUN nimble install -y -d:strip
+
+COPY . .
+RUN nimble build -d:danger -d:lto \
```
zedeus, line 13: `-d:strip` only does something when building.

```suggestion
RUN nimble install -y

COPY . .
RUN nimble build -d:danger -d:lto -d:strip \
```

zedeus:
Thanks, I'm not familiar enough with Docker to really know what's going on here. Is the cache enabled because installing dependencies is a separate step? My first guess would be the step is turned into an image layer, or something.. Some documentation explaining this would be great. Thank you!

zedeus:
`nimble install` downloads dependencies, but it also builds the binary, meaning this will build Nitter twice. To fix this I've added a line to the nimble file so you can do `nimble install -d:depsOnly` to avoid building the binary.

WaywardHeart:
Moving `-d:strip` from build to install is definitely incorrect, but running `nimble install -d` or `nimble install --depsOnly` already skips building the binary, so that change is unnecessary.

zedeus:
Oh yeah, completely missed that Nimble already has this built-in. I'll remove the `depsOnly` define check.

zedeus:
Merged with a few changes, thank you!

Issue #537, opened 2022-02-05
Sometimes footers overlap in embeds


FreeTrade:
Sometimes the footer (date, likes etc) is displayed on top of content when embed is in an iframe - 

example -

https://member.cash/#thread?root=-3065803960502601068&post=-3065803960502601068

PR #538, opened 2022-02-05
Embed CSS tweak

LainLayer:
Attempt to fix https://github.com/zedeus/nitter/issues/537

It seems the image height being set to 100% caused some weird things with the footer in card tweets.

I tested it on a few different types of tweets and nothing seemed broken.

image

I have no idea how to fix this background issue though.


zedeus:
Thanks!

Issue #458, opened 2021-11-04
all pic.jpg_name=orig - fix it please


MiTereKun:
yeah, the site wants to save the pictures not in jpg, but in .jpg_name=orig and every time to delete this crap - it's just annoying. Yes, without this prefix, not the original image opens, but after all, you need to change not even the %addition in the file name%, but to standardize the file format. If you can make a window for uploading files to the site in certain formats, then you can probably do the same for the pictures that the user wants to download.

acarasimon96:
Good catch! This is definitely annoying, and it's something that can be fixed rather quickly. Unfortunately the project has pretty much stalled, so it will take a while before this gets fixed. I would love to work on a PR that fixes this, but right now I'm busy with school.

zedeus:
Not gonna get "fixed" as it's intentional to serve the full resolution.

acarasimon96:
I don't think the submitter meant by be able to change the resolution of the media, and I do like the fact Nitter always serves media in their full resolution. Instead, They were saying that they're having an issue where media would be downloaded with an unrecognized file extension. So, for example, images in JPG format would be saved with a `.jpg_name=orig` extension instead of just `.jpg`.

I'm assuming they're using Windows with file extensions hidden in File Explorer, but nevertheless, it's would be inconvenient (and tedious if they have a lot to save on their devices) for any user to manually change file extensions in the save dialog or in their favorite file manager.

zedeus:
I know, but the only solution I'm aware is removing the resolution query. This is a browser issue, not related to nitter.

MiTereKun:
browser Pale Moon saves in normal resolution pic.jpg_name=orig<b>.jpg</b>. But the browsers based on the chromium do not know how, including mobile opera
one of the sites on which I constantly unload pictures does not understand files
<a><img src="//images.vfl.ru/ii/1636593938/ecc54384/36638325_m.jpg" alt="хнык" title="хнык" border="0"></a>

From solution options - create a JS script that users will need to be downloaded separately, Make a link in about.... And so it would be nice to create a script that over all browsers changes the references to Nitter.net (or the user analog) and Piped

zedeus:
Fixed

PR #539, opened 2022-02-08
Add new endpoint for original resolution images

decoy-walrus:
This change is to work around the issue that chromium based browsers have handling the "name=orig" parameter appended to URLs. This parameter is needed to retrieve the full resolution image from twitter, but causes those browsers to fill in "jpg_name=orig" as the extension on the filename.

This change adds a new endpoint, "/pic/orig/<encoded media>". This new endpoint will internally fetch the URL with ":orig" appended on the end for the full res image. Externally, the endpoint will serve the image without the extra parameter to expose the real extension to the browser.

This new endpoint is used when rendering tweets with attached images. The old endpoint is still in place for all other proxied images, and for any legacy links.

I also updated the "?name=small" parameter to ":small" since that seems to be the new pattern for image sizing.

This should fix issue #458.

### DIFF
--- a/src/routes/media.nim
+++ b/src/routes/media.nim
@@ -88,6 +88,20 @@ proc createMediaRouter*(cfg: Config) =
     get "/pic/?":
       resp Http404
 
+    get re"^\/pic\/orig\/(enc)?\/?(.+)":
+      var url = decoded(request, 1)
+      if "twimg.com" notin url:
+        url.insert(twimg)
+      if not url.startsWith(https):
+        url.insert(https)
+      url.add(":orig")
+
+      let uri = parseUri(url)
+      cond isTwitterUrl(uri) == true
+
+      let code = await proxyMedia(request, url)
+      check code
+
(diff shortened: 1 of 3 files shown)


zedeus:
Thank you, nice solution! I'll do some thorough explorative testing before merging, otherwise it looks good to me. My only concern is going from the `name` param to `:size`, it's hard to know if it will cause any rare issues due to how inconsistent Twitter is.

decoy-walrus:
I've had the code running on my instance for about a day now without any issues, but my user count is low (just me).

I dug up their documentation on how the image URLs should be formatted: https://developer.twitter.com/en/docs/twitter-api/v1/data-dictionary/object-model/entities#photo_format I'm assuming their API docs should still apply to the results we get from the "unofficial" API 🤞 

Looks like you're right, that the size should be a query parameter instead of the ":" colon syntax. I'll play around with that a bit and update this PR.

zedeus:
Works very well, thank you!

PR #548, opened 2022-02-11
Fix RSS <source> tag

niladic:
There is a typo in the RSS view, `<source>` tag has no `>`.

Also [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source) about source tag:
> It must have a start tag, but must not have an end tag.

So this PR also removes `</source>`.


zedeus:
Good catch, thank you!

PR #555, opened 2022-02-19
CSS Tweak to fix avatar issues when "hide tweet stats" is enabled

yoshiyoshyosh:
When using the "hide tweet stats" option, single-line tweets would cut off too early.
image
Removing the position: absolute tag from the avatar fixes this, and from browsing around on both Pale Moon and Chromium with this change, it does not visually inhibit any other scenario.
image

### DIFF
--- a/src/sass/tweet/_base.scss
+++ b/src/sass/tweet/_base.scss
@@ -98,8 +98,6 @@
 }
 
 .avatar {
-    position: absolute;
-
     &.round {
         border-radius: 50%;
     }
--- a/src/sass/tweet/_base.scss
+++ b/src/sass/tweet/_base.scss
@@ -121,11 +121,22 @@
     .tweet-content { 
         font-size: 18px;
     }
+    
+    .tweet-body {
+        display: flex;
+        flex-direction: column;
+        max-height: calc(100vh - 0.75em * 2);
+    }
 
     .card-image img {
         height: auto;
     }
+    
+    .avatar {
+        position: absolute;
+    }
 }
+
 
 .attribution {
     display: flex;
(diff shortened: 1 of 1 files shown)


zedeus:
Unfortunately this breaks embeds.
Normal view:
image

Embed:
image

You can test by adding /embed to the end of any tweet url, e.g. https://nitter.net/elonmusk/status/1494821520815779843/embed

yoshiyoshyosh:
image
Removing the `display: flex` and related CSS appears to have fixed this for embeds.

zedeus:
The lines you removed now were very carefully added. 

Example: https://nitter.net/SpaceX/status/1489252598334984195/embed

Before:
image


After:
image

yoshiyoshyosh:
Ah, my mistake. Backtracking one change, moving the `position: absolute` to `.tweet-embed .avatar` fixes this for the embed layout (A.K.A. keeping it as it was, before any change) while fixing the clipping issues for the normal view (which isn't limited by embed bounds, nor expected to be used as an embed)

zedeus:
Sounds good, I think I added it to the general tweet-body because it seemed harmless. My only concern here is how it affects the mini avatar in media attributions and quotes. You can test with these tweets
Attribution: https://nitter.net/NASAPersevere/status/1476238976335835141
Quote: https://nitter.net/NASAPersevere/status/1494423914507816961#m

yoshiyoshyosh:
> My only concern here is how it affects the mini avatar in media attributions and quotes
Because `.avatar.mini` contains `position: unset`, there is no change at all to small avatars in quotes and attributions.

zedeus:
Thank you! I'll do some testing tomorrow and merge it if everything seems good

PR #557, opened 2022-02-20
Fix bio hashtag

### LINKED ISSUE #521: Encoding problems in bio

### LINKED ISSUE #530: Hashtags written with / without spaces broken
Nitter does not support hashtags separated with only `/` (without extra spaces) — on Twitter it supported instead:

Example:
- `#Hashtag1/#Hashtag2`
- `#Hashtag1/#Hashtag2/#Hashtag3`

See profile about text of:

- https://nitter.net/FreeCADNews
- https://nitter.net/SolveSpace

jackyzy823:
1. The first commit fix #521 and https://github.com/zedeus/nitter/issues/530#issuecomment-1024725110

    For #529 the modification is correct but in wrong file. should modify `src/experimental/parser/user.nim`.

    For https://github.com/zedeus/nitter/issues/530#issuecomment-1024725110  we use std/nre to support Unicode hashtag

    When using nre with `(*U)` , PR #529 is not necessary since regex pattern is generated correctly to treat  `＃` as a single word. and this pattern will also match all Unicode in hashtags.

2. The second commit fix #530 
  remove `/` in htRegex and add negative lookahead to guarantee that URL fragments like `<a href="http://example.com/#testabcd">example.com/#testabcd</a> ` in bio will not be parsed to hashtag .
 also make `#aaa#bbb` behaves the same as twitter which is that these two words are not parsed as hashtag.

### DIFF
--- a/src/experimental/parser/user.nim
+++ b/src/experimental/parser/user.nim
@@ -8,7 +8,7 @@ let
   unRegex = re.re"(^|[^A-z0-9-_./?])@([A-z0-9_]{1,15})"
   unReplace = "$1<a href=\"/$2\">@$2</a>"
 
-  htRegex = nre.re"(*U)(^|[^\w-_./?])([#＃$])([\w_]+)"
+  htRegex = nre.re"""(*U)(^|[^\w-_.?])([#＃$])([\w_]*+)(?!</a>|">|#)"""
   htReplace = "$1<a href=\"/search?q=%23$3\">$2$3</a>"
 
 proc expandUserEntities(user: var User; raw: RawUser) =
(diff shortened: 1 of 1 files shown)


zedeus:
Seems like a decent solution, thanks!

PR #558, opened 2022-02-20
ignore banner url startwith hash in link preload

jackyzy823:
Just to avoid a 404 request like `https://nitter.net/pic/%235bbedc`


zedeus:
Thanks!

PR #562, opened 2022-02-25
Prefer mp4 to m3u8 for Video Playback (for no-js, no-proxy)

girst:
This allows video playback without Javascript and when proxying videos
is turned off.

---

I'm not sure why `playbackType: mp4` was commented out; the default value is m3u8, not mp4, so keeping it as a comment is quite misleading.

Right now, this just takes the first (`[0]`) mp4 url. The mp4 URLs actually contain the resolution of the video file, so it would be good to take the highest quality one. AFAICT they aren't always sorted best-to-worst.

### DIFF
--- a/src/views/tweet.nim
+++ b/src/views/tweet.nim
@@ -100,7 +100,7 @@ proc renderVideo*(video: Video; prefs: Prefs; path: string): VNode =
           renderVideoDisabled(video, path)
         else:
           let vid = video.variants.filterIt(it.contentType == video.playbackType)
-          let source = getVidUrl(vid[0].url)
+          let source = if prefs.proxyVideos: getVidUrl(vid[0].url) else: vid[0].url
           case video.playbackType
           of mp4:
             if prefs.muteVideos:
--- a/src/parser.nim
+++ b/src/parser.nim
@@ -76,8 +76,8 @@ proc parseVideo(js: JsonNode): Video =
     views: js{"ext", "mediaStats", "r", "ok", "viewCount"}.getStr,
     available: js{"ext_media_availability", "status"}.getStr == "available",
     title: js{"ext_alt_text"}.getStr,
-    durationMs: js{"video_info", "duration_millis"}.getInt
-    # playbackType: mp4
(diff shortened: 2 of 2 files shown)


zedeus:
The reason I haven't done this yet is the entire file will be downloaded and sent to the browser in 1 chunk instead of streamed. This is especially a problem for larger videos. I have an experimental solution I made a while ago for Nitter to support mp4 streaming, but it had some edge cases that resulted in videos not playing at all occasionally. I appreciate your PR, but it's going to cause quite excessive memory usage for even smaller instances. The experimental solution may become a PR soon-ish if I can get it working again.

src/views/tweet.nim
```diff
@@ -100,7 +100,7 @@ proc renderVideo*(video: Video; prefs: Prefs; path: string): VNode =
           renderVideoDisabled(video, path)
         else:
           let vid = video.variants.filterIt(it.contentType == video.playbackType)
```
girst, line 102:
```suggestion
          let playbackType = if prefs.proxyVideos: video.playbackType else: mp4
          let vid = video.variants.filterIt(it.contentType == video.playbackType)
```

girst (commented):
>The reason I haven't done this yet is the entire file will be downloaded and sent to the browser in 1 chunk instead of streamed. This is especially a problem for larger videos. 

That's a valid concern. What do you think about the approach in this self-review? essentially, keep 
playbackType the default in parser.nim and only override it if proxyVideos is false. That way, mp4 is only used in the case where using m3u8 will not work.

>I have an experimental solution I made a while ago for Nitter to support mp4 streaming, but it had some edge cases that resulted in videos not playing at all occasionally. [...] The experimental solution may become a PR soon-ish if I can get it working again.

That would be great, too!

src/views/tweet.nim
zedeus: This could work, my concern here would be in the odd case where mp4 isn't available, which I think is mostly the case for "pro" videos (ads and ad-like business posts) and maybe some very old videos. A fallback would be necessary, so this bit of code should be able to handle cases where mp4 isn't in the list of variants.

PR #563, opened 2022-02-25
[v2] Prefer mp4 to m3u8 for Video Playback

girst:
Implements the approach outlined in #562. the mp4 video is only used if it actually exists, so no problems should arise with "pro" videos/business posts.

also now sorts mp4s by resolution (height of video). i couldn't find a json key specifying the resolution, so i had to parse it out of the url.

### DIFF
--- a/src/parser.nim
+++ b/src/parser.nim
@@ -87,10 +87,15 @@ proc parseVideo(js: JsonNode): Video =
     result.description = description.getStr
 
   for v in js{"video_info", "variants"}:
+    let
+      contentType = parseEnum[VideoType](v{"content_type"}.getStr("summary"))
+      url = v{"url"}.getStr
+      resolution = getMp4Resolution(url)  # only available if contentType == mp4
     result.variants.add VideoVariant(
-      contentType: parseEnum[VideoType](v{"content_type"}.getStr("summary")),
+      contentType: contentType,
       bitrate: v{"bitrate"}.getInt,
-      url: v{"url"}.getStr
+      url: url,
+      resolution: resolution
     )
 
 proc parsePromoVideo(js: JsonNode): Video =
--- a/src/views/tweet.nim
+++ b/src/views/tweet.nim
@@ -62,10 +62,13 @@ proc renderAlbum(tweet: Tweet): VNode =
             a(href=getPicUrl(orig), class="still-image", target="_blank"):
               genImg(small)
 
-proc isPlaybackEnabled(prefs: Prefs; video: Video): bool =
-  case video.playbackType
+proc isPlaybackEnabled(prefs: Prefs; playbackType: VideoType): bool =
+  case playbackType
   of mp4: prefs.mp4Playback
   of m3u8, vmap: prefs.hlsPlayback
+
+proc hasMp4Url(video: Video): bool =
+  video.variants.anyIt(it.contentType == mp4)
 
 proc renderVideoDisabled(video: Video; path: string): VNode =
   buildHtml(tdiv(class="video-overlay")):
@@ -87,6 +90,9 @@ proc renderVideo*(video: Video; prefs: Prefs; path: string): VNode =
   let container =
     if video.description.len > 0 or video.title.len > 0: " card-container"
     else: ""
+  let playbackType =
+    if not prefs.proxyVideos and video.hasMp4Url: mp4
+    else: video.playbackType
 
   buildHtml(tdiv(class="attachments card")):
     tdiv(class="gallery-video" & container):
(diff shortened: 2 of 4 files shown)


src/views/tweet.nim
```diff
@@ -67,6 +67,9 @@ proc isPlaybackEnabled(prefs: Prefs; video: Video): bool =
   of mp4: prefs.mp4Playback
   of m3u8, vmap: prefs.hlsPlayback
 
+proc hasMp4Url(video: Video): bool =
+  (video.variants.filterIt(it.contentType == mp4)).len > 0
```
zedeus, line 71:
```suggestion
  video.variants.anyIt(it.contentType == mp4)
```

src/views/tweet.nim
```diff
@@ -99,9 +108,10 @@ proc renderVideo*(video: Video; prefs: Prefs; path: string): VNode =
           img(src=thumb)
           renderVideoDisabled(video, path)
         else:
-          let vid = video.variants.filterIt(it.contentType == video.playbackType)
-          let source = getVidUrl(vid[0].url)
-          case video.playbackType
+          let vid = video.variants.filterIt(it.contentType == playbackType)
+          let vidUrl = sorted(vid, cmpVideosByResolution, Descending)[0].url
+          let source = if prefs.proxyVideos: getVidUrl(vidUrl) else: vidUrl
```
zedeus, line 113:
```suggestion
          let
            vars = video.variants.filterIt(it.contentType == playbackType)
            vidUrl = vars.sortedByIt(it.resolution).reversed[0].url
            source = if prefs.proxyVideos: getVidUrl(vidUrl)
                     else: vidUrl
```

With these changes you can remove the `cmpVideosByResolution` function

src/views/tweet.nim
```diff
@@ -83,10 +86,16 @@ proc renderVideoUnavailable(video: Video): VNode =
...
+proc cmpVideosByResolution(x, y: VideoVariant): int =
+  cmp(x.resolution, y.resolution)
+
 proc renderVideo*(video: Video; prefs: Prefs; path: string): VNode =
   let container =
     if video.description.len > 0 or video.title.len > 0: " card-container"
     else: ""
+  let playbackType =
+    if not prefs.proxyVideos and hasMp4Url(video): mp4
+    else: video.playbackType
```
zedeus, line 98:
```suggestion

  let playbackType =
    if not prefs.proxyVideos and video.hasMp4Url: mp4
    else: video.playbackType
```

zedeus (changes requested):
Thanks! Just a few stylistic changes

girst:
great feedback, thanks! i guess it's pretty obvious i'm not fluent in nim :D

girst:
Note: the [latest force-push](https://github.com/zedeus/nitter/compare/dcae3b198849e07d0a0c4cdd900bb4fa82a273c8..0efe5967c4e821c3d24a9a0043fa3f2fbbf9c10f) fixes the bug where videos loaded as mp4 still showed the "enable HLS" button.

src/views/tweet.nim
```diff
@@ -95,13 +101,16 @@ proc renderVideo*(video: Video; prefs: Prefs; path: string): VNode =
...
+        elif not prefs.isPlaybackEnabled(playbackType):
           img(src=thumb)
           renderVideoDisabled(video, path)
         else:
-          let vid = video.variants.filterIt(it.contentType == video.playbackType)
-          let source = getVidUrl(vid[0].url)
-          case video.playbackType
+          let
+            vars = video.variants.filterIt(it.contentType == playbackType)
+            vidUrl = vars.sortedByIt(it.resolution).reversed[0].url
```
zedeus, line 110: `.reversed[0]` is equivalent to `[^1]` using a [BackwardsIndex](https://nim-lang.org/docs/system.html#^.t,int) to get the last element

src/parser.nim
```diff
@@ -87,10 +87,16 @@ proc parseVideo(js: JsonNode): Video =
     result.description = description.getStr
 
   for v in js{"video_info", "variants"}:
+    let contentType = parseEnum[VideoType](v{"content_type"}.getStr("summary"))
+    let url = v{"url"}.getStr
+    let resolution =
+      if contentType == mp4 and url =~ re".*/vid/\d+x(\d+)/": parseInt(matches[0])
+      else: 0
```
zedeus, line 94:
```suggestion
    let 
      contentType = parseEnum[VideoType](v{"content_type"}.getStr("summary"))
      url = v{"url"}.getStr
      resolution =
        if contentType == mp4 and url =~ re".*/vid/\d+x(\d+)/": parseInt(matches[0])
        else: 0
```

src/views/tweet.nim
zedeus, line 110: `.reversed[0]` is equivalent to `[^1]` using a [BackwardsIndex](https://nim-lang.org/docs/system.html#^.t,int) to get the last element

```suggestion
            vidUrl = vars.sortedByIt(it.resolution)[^1].url
```

zedeus (changes requested):
Sorry for the late review, just a couple nits left.

src/parser.nim
```diff
@@ -87,10 +87,16 @@ proc parseVideo(js: JsonNode): Video =
     result.description = description.getStr
 
   for v in js{"video_info", "variants"}:
+    let contentType = parseEnum[VideoType](v{"content_type"}.getStr("summary"))
+    let url = v{"url"}.getStr
+    let resolution =
+      if contentType == mp4 and url =~ re".*/vid/\d+x(\d+)/": parseInt(matches[0])
```
zedeus, line 93: This could be turned into a `getMp4Resolution` proc in `parserutils.nim`, using a few `find` calls instead of regex, making it orders of magnitude more efficient. Here's an equivalent function tested with a few URLs:

```nim
proc getMp4Resolution(url: string): int =
  const vidSep = "/vid/"
  let
    vidIdx = url.find(vidSep) + vidSep.len
    resIdx = url.find('x', vidIdx) + 1
    res = url[resIdx ..< url.find("/", resIdx)]

  if res.len > 0:
    return parseInt(res)
```

girst:
all feedback applied and tested again. i modified your `getMp4Resolution` so it can also handle non-conforming URLs, making it safe to call on all contentTypes.

zedeus:
Merged with some minor changes, thank you!

PR #564, opened 2022-02-26
Add "Source" to main tweets

110Percent:
Add the name of the Twitter client used to send a tweet when rendering a main tweet, similar to the Twitter website.

image

### DIFF
--- a/src/parser.nim
+++ b/src/parser.nim
@@ -189,7 +189,7 @@ proc parseTweet(js: JsonNode): Tweet =
     replyId: js{"in_reply_to_status_id_str"}.getId,
     text: js{"full_text"}.getStr,
     time: js{"created_at"}.getTime,
-    source: js{"source"}.getStr.split(">")[1].split("<")[0],
+    source: getSource(js),
     hasThread: js{"self_thread"}.notNull,
     available: true,
     user: User(id: js{"user_id_str"}.getStr),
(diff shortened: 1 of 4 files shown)


src/parser.nim
```diff
@@ -189,6 +190,7 @@ proc parseTweet(js: JsonNode): Tweet =
     replyId: js{"in_reply_to_status_id_str"}.getId,
     text: js{"full_text"}.getStr,
     time: js{"created_at"}.getTime,
+    source: js{"source"}.getStr.split(">")[1].split("<")[0],
```
zedeus, line 193: Just in case source ever changes format or is empty we need to extract it in a way that won't crash on errors:
```nim
str.substr(str.find('>') + 1, str.rfind('<') - 1)
```

You can put this in a `getSource` proc in parserutils below `getTombstone`

src/types.nim
```diff
@@ -175,6 +175,7 @@ type
     tombstone*: string
     location*: string
     stats*: TweetStats
+    source*: string
```
zedeus, line 178: Minor nit, put this above stats

zedeus (changes requested):

110Percent:
Good ideas; I'm not familiar with Nim, so a lot of this is foreign to me. I've pushed the requested changes.

zedeus:
Merged manually with a few fixes. Thanks!

PR #566, opened 2022-03-04
Implement healthcheck on nitter and redis

dennis1248:
This will implement a basic healthcheck in the docker-compose file for both Nitter and Redis. The healthcheck can be used by administrators to easily spot potential problems with a running container. It can also be used to automatically restart or recreate unhealthy containers (eg. via willfarrell/docker-autoheal) should the administrator desire to do so.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,11 @@ services:
     depends_on:
       - nitter-redis
     restart: unless-stopped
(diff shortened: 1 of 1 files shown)


dennis1248:
Removed and accidentally added empty line.

zedeus:
Thanks!

PR #571, opened 2022-03-10
 Fix list members not displaying

jackyzy823:
with empty cursor , twitter raise 'Dependency: Unspecified' error. so we add curosr only when it is not empty.


zedeus:
Thanks!

Issue #76, opened 2019-10-14
Add RTL (bidi) support


ahangarha:
There should be support for RTL languages such as Persian, Arabic, Urdu,...

Currently, all the tweet texts are rendered with LTR direction. This can be done by applying bidi to the project.

![image](https://user-images.githubusercontent.com/11241315/66772659-14b93f80-eeca-11e9-9a95-db925859a313.png)
you may see the current status of Nitter that cannot render RTL text in RTL direction. The first tweet, I have applied the following explanation while the second tweet is following the default Nitter layout.

In quick review, I found that there is a need to add `dir="auto"` to the `div` containing the tweet text and remove the `display: inline` css property from it. I think this would solve the issue but for sure it needs more in depth investigation.

zedeus:
Thank you! I've applied your suggestion and it seems to be working properly.

zedeus:
I spent some time researching the best way to solve this, but unfortunately supporting bidirectional text by default makes clicking tweets quite annoying. The entire text box (including the empty part) blocks clicks, and we don't want `pointer-events: none` since you wouldn't be able to select the text. For these reasons I've put bidi support behind an opt-in preference. If you can come up with a solution that support bidirectional support without blocking click events, I'd be happy to merge it.

ahangarha:
I don't face with any issue!

rezad1393:
this bug is still present
example
https://nitter.pussthecat.org/general_g_e_m
and it is shown correctly in twitter.

rezad1393:
> @rezad1393 did you enable the bidirectional text preference?

didnt see that .
thanks.

PR #605, opened 2022-04-09
Make tweet-content inline-block to get direction

ahangarha:
Currently the body of tweet has `display: inline;` style. This, prevents getting proper direction and alignment when the language doesn't follow the global language in terms of direction (LTR/RTL).

image

After this change (which shouldn't be breaking change), the text gets rendered in correct direction and alignment.

### DIFF
--- a/src/sass/tweet/_base.scss
+++ b/src/sass/tweet/_base.scss
@@ -20,7 +20,7 @@
     font-family: $font_3;
     line-height: 1.3em;
     pointer-events: all;
-    display: inline;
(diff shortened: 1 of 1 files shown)


zedeus:
This results in the same issue I already mentioned when you created an issue (https://github.com/zedeus/nitter/issues/76#issuecomment-723509162). If you want RTL support, enable it on the preferences page.

ahangarha:
I am sorry. I forgot such issues was discussed before.

PR #623, opened 2022-05-16
Video embeds

HookedBehemoth:
This was made in response to TwitFix/fxtwitter shutting down.

It uses the twitter:player:* meta tags with mp4 video links.

The big downside right now is that descriptions won't be embedded on videos. TwitFix worked around this with yet another embed tag called [oEmbed](https://oembed.com/).

sample
image

### DIFF
--- a/src/routes/status.nim
+++ b/src/routes/status.nim
@@ -46,8 +46,10 @@ proc createStatusRouter*(cfg: Config) =
         video = ""
 
       if conv.tweet.video.isSome():
-        images = @[get(conv.tweet.video).thumb]
-        video = getVideoEmbed(cfg, conv.tweet.id)
+        let vid = get(conv.tweet.video)
+        images = @[vid.thumb]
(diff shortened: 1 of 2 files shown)


zedeus:
Very interesting solution, thanks! Unfortunately it's not ready yet due to the issue mentioned here: https://github.com/zedeus/nitter/pull/562#issuecomment-1051280759

HookedBehemoth:
Digging through the proxy code, the actual data transfer code appears to already be chunked [proxyMedia](https://github.com/zedeus/nitter/blob/138826fb4fbdec73fc6fee2e025fda88f7f2fb49/src/routes/media.nim#L63-L65), no?

zedeus:
Chunked yes, but special support for Twitter's mp4 quirks was necessary. Implementation here, testing would be greatly appreciated: #634

HookedBehemoth:
I gave up. Discord keeps breaking embeds.

Issue #472, opened 2021-12-01
Put rel=canonical links in page headers to redirect search engine attention


gThorondorsen:
Continuing the conversation in #431 brought up one point that I think could be actioned: indexing of Nitter instances by search engines hurts SEO of the original tweet. Conversely, some (most?) instance administrators could do without the attention that brings to their domain. (I know of instances that closed because they did not have the capacity to deal with the stream of #431-like complaints they got.)

The `rel="canonical"` link type has been designed by search engines implementers[[RFC6596](https://datatracker.ietf.org/doc/html/rfc6596)] for exactly such a purpose: telling crawlers that the page is a duplicate of another resource. It does work cross-domain, though I do not know how well that case is supported by search engines. Of the other site-specific proxying services that I know of, only [Invidious](https://github.com/iv-org/invidious) does something similar: it uses a `rel="alternate"` link [as a hint to `youtube-dl`](https://github.com/iv-org/invidious/pull/1730) (and forks). That may also work for search engines but feels like a weaker hint that might not have the desired effect. Someone with more SEO experience should chime in.

The `robots.txt` file should workaround this issue (and some more), but this currently relies on every instance administrator getting it right and not assuming Nitter comes bundled with it. Also, it is my understanding that such a solution *removes* SEO weight altogether when someone links to a Nitter instance while the `rel="canonical"` one *transfers* it to the twitter.com address.

Regarding the implementation, there are two points to watch out for.

- There are a few page that could be excluded from `rel="canonical"`, though `rel="alternate"` could apply: `/` and `/about` are obvious candidates, `/settings` feels like a subtle case worth discussing in details, and I have no idea about `/search`.

- What flexibility should instance administrators have? And what should be the default? I've currently settled weakly on `rel="canonical"` being the default, with instances being able to change (downgrade?) it to `rel="alternate"` or nothing if they feel confident in their `robots.txt` settings and/or want the search engine attention.

I am willing to do implementation work if needed, but I'll first wait for approval on the general direction and feedback on the pending design questions before seriously diving into it.

arrowgent:
the point of privacy focused nitter is to remove tracking
rel="x" is url tracking.

gThorondorsen:
> rel="x" is url tracking.

[...]

Issue #470, opened 2021-11-22
Problem with the number of requests from GoogleBot


D3m1tx:
Hi,

I have a problem with google bots as they have multiplied x4 the cost of my nitter instances in Google Cloud (from 0.8$/day to 4$/day). Half of the requests come from googlebot.

Does anyone know how to add the robots.txt file easily starting from the precompiled docker image?
I have tried to pull the image, add the robots.txt to the container (docker cp...) and create a new image containing robots.txt, but when I deploy it to Google Cloud I get an error
`./start.sh: line 4: ./nitter: not found`
Both images seem to be the same size. Does anyone know how to fix this, or does anyone have precompiled an image that prevents crawling?

Thanks,

jackyzy823:
Create your `robots.txt` and then do as follow:

` docker run  -p 8080:8080 -v `pwd`/nitter.conf:/src/nitter.conf -v `pwd`/robots.txt:/src/public/robots.txt zedeus/nitter`

D3m1tx:
Hi jackyzy823,

Thank you for your reply. Although what you comment works when you run the container locally, when trying to create an image from that container, the volumes are not copied, so the robots.txt information is lost.

Google Cloud Run does not allow using docker volumes and therefore I have to generate a docker image with the robots.txt already included.

I will keep trying with the copy and backup volumes commands.

Thanks.

jackyzy823:
Dockerfile like below and robots.txt in the same folder
```Dockerfile
FROM zedeus/nitter
MAINTAINER yourself
COPY robots.txt /src/public
```
do 
`docker build . -t <IMAGE_TAG>`

D3m1tx:
It was easier than I thought.

`docker run -p 8080:8080`
`docker cp LOCAL_PATH_TO_ROBOTS_FILE CONTAINER_NAME:/src/public/robots.txt`
`docker commit CONTAINER_NAME zedeus/nitter`
`docker push zedeus/nitter`
You must replace zedeus/nitter for your repository

I had the problem because when using the command
docker run -p 8080:8080 -v `pwd`/nitter.conf:/src/nitter.conf -v `pwd`/robots.txt:/src/public/robots.txt zedeus/nitter
and then try to copy the file with the cp command, I got an error because it was already in use.

Sorry for the inconvenience and thank your for your reply.

D3m1tx:
> Dockerfile like below and robots.txt in the same folder
> 
> ```dockerfile
> FROM zedeus/nitter
[...]

Issue #305, opened 2021-01-09
Instance has been rate limited


stemy2:
Hello,

According the readme, there is no rate limits, but since not long ago, i have often the message 

```
Instance has been rate limited.
Use another instance or try again later.
```

I'd like to know why there is actually rate limits even if the readme says there isn't. Did twitter changed something ?

girst:
i was wondering the same. it looks like, according to this commit (https://github.com/zedeus/nitter/commit/4e1d21348844d023eada9c8161021f322fedb432), that the rate-limiting is coming from twitter's side, not nitter's.

the error message is a bit ambiguous; maybe @zedeus can add "... by twitter" to it?

Mennaruuk:
I've been noticing this error primarily on Nitter.net. It seems like other instances aren't facing the same issue. I'm currently using [FDNitter ](https://nitter.fdn.fr)and it hasn't been rate limited.

stemy2:
Nittet.net is not the only instance affected. Maybe an automated redirect to another instance wgen limit is reached could be a good idea.

zedeus:
Automatically redirecting will just make DDOSing all instances at the same time easier. I'll work on ways to bypass the rate limits instead.

FireMasterK:
Can we add a robots.txt file to further block crawlers?

FireMasterK:
I'm not sure if this helps:

My instance used to say rate limited when loading pages, but after reloading the page once it loaded fine

When I restarted my instance, the problem went away entirely.

So, this raises the question - are these ratelimits per token? Can't we just discard rate-limited tokens and fetch new ones?

Mennaruuk:
I remember zedeus mentioning last year that it was "[500 tokens per 30 minutes](https://github.com/zedeus/nitter/issues/200#issuecomment-648813934)", I don't know if that still holds true today with the changes to Twitter API v2.

zedeus:
From a local test, it seems they actually increased it to 1000 tokens per 30 minutes, which is nice, but apparently not enough for Nitter. I don't understand that though considering that would result in a theoretical max of at least 1 million requests per 15 minutes (I won't get into the math, but it has to do with the rate limit windows and token expiration time). I'm fairly certain my half-assed token pool is the problem, I think it throws away tokens that hit the 15-minute window rate limit instead of saving them for later, but I could be wrong. Some simple logging showed only 19 tokens are in the pool when it hits the rate limit, which makes no sense considering the max of 1000 per 30 minutes.

[...]

PR #631, opened 2022-06-04
Block search engines via robots.txt

minus7:
Prevents instances from being rate limited due to being senselessly
crawled by search engines. Since there is no reason to index Nitter
instances, simply block all robots. Notably, this does *not* affect link
previews (e.g. in various chat software).

Invidious and Bibliogram do already ship a robots.txt by default and I think
it's a good thing for Nitter to do the same, looking at #305, #470 and that
rel=canonical (#472) seems to do nothing.

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /


zedeus:
Thanks!

PR #659, opened 2022-07-02
Reintroduce space around delimiter in html title for better readability

Nachtalb:
The `|` delimiter was surrounded by spaces before 21e8f04fa47cbef5332210eb650ef7819cf9bca8 which I believe was removed by accident. The spaces allow the text to break into a new line in certain situations and so on which improves readability.


zedeus:
It was indeed by accident, but it turns out my usage of strformat is a bit slower than simple concats, so the commit will be somewhat reverted. I'll keep this open as a reminder until it's done, thanks for spotting it!

zedeus:
Fixed in https://github.com/zedeus/nitter/commit/85316f8f8d265ea4449e7166c1f6651fb595c41d

PR #669, opened 2022-07-17
fix profile-website css

jackyzy823:
just a typo fix.

### DIFF
--- a/src/sass/profile/card.scss
+++ b/src/sass/profile/card.scss
@@ -73,9 +73,9 @@
         }
     }
 
-    .profile-joindate, .profile-location, profile-website {
+    .profile-joindate, .profile-location, .profile-website {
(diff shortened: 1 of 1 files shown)


src/sass/profile/card.scss
```diff
@@ -73,7 +73,7 @@
         }
     }
 
-    .profile-joindate, .profile-location, profile-website {
+    .profile-joindate, .profile-location, .profile-website {
         color: var(--fg_faded);
         margin: 2px 0;
```
zedeus, line 78: With this change the section gets slightly bigger, remedied by this:
```suggestion
        margin: 1px 0;
```

jackyzy823: done.

zedeus:
Thanks, good catch

PR #694, opened 2022-09-15
bump karax version

ringabout:
My bad, but the karax dep needs to include https://github.com/karaxnim/karax/pull/230 (patches for ARC/ORC) too.


ringabout:
Thank you!

PR #735, opened 2022-11-25
Change :ro to :Z

TommyTran732:
:Z is needed for SELinux systems. The side effect of setting :Z is that the config will be writable inside of the container, but I do not think it matters too much.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
     ports:
       - "127.0.0.1:8080:8080" # Replace with "8080:8080" if you don't use a reverse proxy
     volumes:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
     ports:
       - "127.0.0.1:8080:8080" # Replace with "8080:8080" if you don't use a reverse proxy
     volumes:
-      - ./nitter.conf:/src/nitter.conf:ro
+      - ./nitter.conf:/src/nitter.conf:Z
```
mikroskeem, line 11:
```suggestion
      - ./nitter.conf:/src/nitter.conf:Z,ro
```

PR #742, opened 2022-12-03
Expand acronym

vanillajonathan:


kokocivikcv:
https://PolygonScan.io/
@kokocivikcv19

zedeus:
It's not for live streaming so that makes it more confusing. It's not important what hls means.

vanillajonathan:
Then it should at least be uppercased so that it is obvious that is an acronym and not a misspelled word.

zedeus:
Done

PR #756, opened 2022-12-31
Bump zippy dependency (fixes #750 arm64 build error)

apsun:
https://github.com/guzba/zippy/commit/e67b880fed12ff095bffdee8d74ee4bf34570151 fixes a type mismatch on arm64 builds.

PR #762, opened 2023-01-06
Fix Dockerfile on arm64

Whizanth:
Now that the issue with zippy has been fixed (#756), the only thing preventing Nitter's Dockerfile to successfully build on arm64 is the Nim image.

To solve this, we can use the regular Alpine image, and install Nim manually.

With the changes in this pull request, I was successfully able to build the Nitter image on both arm64 and x86_64.

### DIFF
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -1,0 +1,23 @@
+FROM alpine:3.17 as nim
+LABEL maintainer="setenforce@protonmail.com"
+
+RUN apk --no-cache add gcc git libc-dev libsass-dev "nim=1.6.8-r0" nimble pcre
+
+WORKDIR /src/nitter
+
+COPY nitter.nimble .
+RUN nimble install -y --depsOnly
+
+COPY . .
+RUN nimble build -d:danger -d:lto -d:strip \
+    && nimble scss \
+    && nimble md
+
+FROM alpine:3.17
+WORKDIR /src/
+RUN apk --no-cache add pcre ca-certificates
+COPY --from=nim /src/nitter/nitter ./
+COPY --from=nim /src/nitter/nitter.example.conf ./nitter.conf
+COPY --from=nim /src/nitter/public ./public
+EXPOSE 8080
+CMD ./nitter


Whizanth:
I've updated my fork to use a separate Dockerfile for arm64, as you suggested.

As for enforcing the Nim version, as far as I'm able to tell, the latest Nim version that's available in the Alpine repositories is 1.6.8-r0. In contrast, the official Dockerfile for x86 uses 1.6.10, which is newer, so I don't think enforcing an older version would be a good idea, tell me if you don't agree.

If all of this is okay, do I need to squash my commits into one before you merge?

unixfox:
> As for enforcing the Nim version, as far as I'm able to tell, the latest Nim version that's available in the Alpine repositories is 1.6.8-r0. In contrast, the official Dockerfile for x86 uses 1.6.10, which is newer, so I don't think enforcing an older version would be a good idea, tell me if you don't agree.

That doesn't really matter, it's to avoid using a newer version if one day alpine edge have a much newer version than the version specified in the official Dockerfile.

zedeus:
Thanks, this looks good. Assuming Docker can automatically use this file to build an ARM image, this should also be reverted: https://github.com/zedeus/nitter/commit/2aa07e7395f370fff6128cab68b5995feb09e156

unixfox:
> Thanks, this looks good. Assuming Docker can automatically use this file to build an ARM image, this should also be reverted: https://github.com/zedeus/nitter/commit/2aa07e7395f370fff6128cab68b5995feb09e156

Here it's a different Dockerfile, it can't be built with the original Dockerfile image.

I think @Whizanth you will need to add a github workflow for pushing the image too.

Dockerfile.arm64
```diff
@@ -0,0 +1,23 @@
+FROM alpine as nim
+LABEL maintainer="setenforce@protonmail.com"
+
+RUN apk --no-cache add gcc git libc-dev libsass-dev "nim=1.6.10" nimble pcre
```
unixfox, line 4: It's `nim=1.6.8-r0` and not `nim=1.6.10`

Dockerfile.arm64
```diff
@@ -0,0 +1,23 @@
+FROM alpine as nim
```
unixfox, line 1: Please tag an alpine release, like `FROM alpine:3.17 as nim`

Dockerfile.arm64
```diff
@@ -0,0 +1,23 @@
...
+
+COPY nitter.nimble .
+RUN nimble install -y --depsOnly
+
+COPY . .
+RUN nimble build -d:danger -d:lto -d:strip \
+    && nimble scss \
+    && nimble md
+
+FROM alpine:latest
```
unixfox, line 16: Same as https://github.com/zedeus/nitter/pull/762/files#r1118086428

unixfox (changes requested):
Please see my comments

Dockerfile.arm64
Whizanth: Resolved.

zedeus:
Thank you!

Whizanth:
You're welcome! The GitHub workflow will still have to be added though, but I'm not familiar with it.

PR #799, opened 2023-02-26
Update note about ARM64 docker image + add docker hub links

unixfox:
So people are aware of this new docker image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,9 @@ performance reasons.
 
 ### Docker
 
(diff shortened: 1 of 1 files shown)


zedeus:
Thanks, while you're at it can you add a link to the Docker hub page? I realized it's missing.

unixfox:
> Thanks, while you're at it can you add a link to the Docker hub page? I realized it's missing.

Done

PR #802, opened 2023-03-01
parse long tweets

HookedBehemoth:
This enables the feature `longform_notetweets_consumption_enabled` in graphql and parses the note tweets if they are available.
Sadly the nature of twitters horrendous API still requires the old routine to parse the "legacy" text to get the correct replies which are not obtainable in `note_tweet`.
<details>

</details>

### DIFF
--- a/src/parser.nim
+++ b/src/parser.nim
@@ -385,6 +385,10 @@ proc parseGraphTweet(js: JsonNode): Tweet =
   result = parseTweet(js{"legacy"}, jsCard)
   result.user = parseUser(js{"core", "user_results", "result", "legacy"})
 
+  var note_tweet = js{"note_tweet", "note_tweet_results", "result"}
+  if note_tweet.kind != JNull:
+    result.expandNoteTweetEntities(note_tweet)
+
   if result.quote.isSome:
     result.quote = some(parseGraphTweet(js{"quoted_status_result", "result"}))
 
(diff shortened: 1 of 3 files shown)


src/parser.nim
```diff
@@ -385,6 +385,10 @@ proc parseGraphTweet(js: JsonNode): Tweet =
   result = parseTweet(js{"legacy"}, jsCard)
   result.user = parseUser(js{"core", "user_results", "result", "legacy"})
 
+  var note_tweet = js{"note_tweet", "note_tweet_results", "result"}
+  if note_tweet.kind != JNull:
+    result.expandNoteTweetEntities(note_tweet)
```
zedeus, line 390:
```suggestion
  with noteTweet, js{"note_tweet", "note_tweet_results", "result"}:
    result.expandNoteTweetEntities(noteTweet)
```

src/parserutils.nim
```diff
@@ -289,3 +289,45 @@ proc expandTweetEntities*(tweet: Tweet; js: JsonNode) =
 
   tweet.text = orig.replacedWith(replacements, textSlice)
                    .strip(leading=false)
+
+proc expandNoteTweetEntities*(tweet: Tweet; noteTweet: JsonNode) =
```
zedeus, line 293: We can probably avoid this, see my message on matrix

zedeus (changes requested):

PR #812, opened 2023-03-21
GraphQL timeline

zedeus:


src/consts.nim
```diff
@@ -19,8 +19,10 @@ const
   tweet* = timelineApi / "conversation"
 
   graphql = api / "graphql"
+  graphUserTweets* = graphql / "9rys0A7w1EyqVd2ME0QCJg/UserTweets"
+  graphUserTweetsAndReplies* = graphql / "ehMCHF3Mkgjsfz_aImqOsg/UserTweetsAndReplies"
   graphTweet* = graphql / "6lWNh96EXDJCXl05SAtn_g/TweetDetail"
-  graphUser* = graphql / "7mjxD3-C6BxitPMVQ6w0-Q/UserByScreenName"
+  graphUser* = graphql / "nZjSkpOpSL5rWyIVdsKeLA/UserByScreenName"
```
markcellus, line 25: May I ask, where are these values coming from? Are you seeing these show up in the network calls? If so, are they ephemeral?

zedeus: It's some sort of versioning ID that gives you persistent access to a specific version of the API. It's probably random, haven't tried to decode it, but you can find plenty of versions by searching for the endpoints on GitHub and elsewhere.

zedeus: You can get the newest one by opening devtools, going to the network tab, filtering by graphql, then loading pages on Twitter.

samtin0x:
Is everything set to merge this PR? thanks!

PR #820, opened 2023-03-29
Include quoted tweet body in RSS

### LINKED ISSUE #132: Include original posting-URL into commenting-retweets - feature request
In nitter's RSS feed the tweet content appears in the Feed entry title. This also is the case with RT: retweets. If a tweet of a news channel or blog is retweetet without adding a comment, the content of the original tweet appears in the RSS feed of the retweeter. OK, but this is not the case, if an own comment was added for the retweet. 

Feature request: Would you please append the original tweet to the comment of the retweeter? Reason: This would bring the posted URL into the title of the RSS feed of the retweeting account.

gil:
Hey folks! This is my quick attempt of displaying quoted tweets on RSS results. [As I've said](https://github.com/zedeus/nitter/issues/132#issuecomment-1486831574), I don't know this codebase nor have worked with Nim before, so I'm sorry for any problems it may cause. So far, it's working well on my test instance and I like the results much more. This is how it looks:

### Before

Screenshot of a tweet with a link to a quoted tweet on an RSS reader

### After

Screenshot of a tweet with a quoted one being displayed on an RSS reader

Fixes #132


ssimpson89:
This would be much appreciated!

woj-tek:
@gil - it looks great, would be awesome to have it included :-)

src/views/rss.nimf
```diff
@@ -54,6 +50,12 @@ Twitter feed for: ${desc}. Generated by ${cfg.hostname}
 <img src="${urlPrefix}${getPicUrl(card.image)}" style="max-width:250px;" />
 #  end if
 #end if
+#if tweet.quote.isSome and get(tweet.quote).available:
+#  let quoteLink = getLink(get(tweet.quote))
+<hr/>
+<p>Quoting: <a href="${urlPrefix}${quoteLink}">${cfg.hostname}${quoteLink}</a></p>
+${renderRssTweet(get(tweet.quote), cfg)}
+#end if
```
ninboy, line 58: Let me start by saying that this is the sole reason it's keeping me away from using Nitter's RSS feeds and I'm grateful you pointed in the right direction here. I hope this get's merged as soon as possible 🤞 

My 2 cents here would be that this should use more semantically correct element: 
```suggestion
<blockquote>
<p>
$renderRssTweet(get(tweet.quote), cfg)}
</p>
<footer>
— <cite><a href="${urlPrefix}${quoteLink}">${cfg.hostname}${quoteLink}</a></cite>
</footer>
</blockquote>
```

koszzz:
I need it, that's wonderful.

koszzz:
I hope this get's merged

src/views/rss.nimf
gil: Thanks for the suggestion! It looks much better than what I did. Sorry I didn't merge it before, I was busy and then Twitter started breaking Nitter. Let's see if things will get more stable soon and I'll go back to this!

gil:
Yeah, me too @koszzz ! But at this point, I'm not even sure if Nitter will ever be stable again so that we can use it for RSS 😢

untitaker:
this patch seems to work for my usecases and is now deployed on `nitter.woodland.cafe`

thomas-333:
I appreciate this is a few months old now but this is something I'm very interested in as I use my local Nitter instance for rss.

When I try and apply manually and compile I'm getting an error:

```
/src/nitter/src/views/rss.nimf(57, 12) Error: type mismatch
Expression: $renderRssTweet
  [1] renderRssTweet: proc (tweet: Tweet, cfg: Config): string
```

If I try and apply automatically the diff fails, presumably the code has moved on since?

```
curl https://github.com/zedeus/nitter/commit/d2b858e5e32107e558b4cecff08f07e13b1b93e2.diff | git apply
error: patch failed: src/views/rss.nimf:54
error: src/views/rss.nimf: patch does not apply
```

gil:
@thomas-333 yeah, I imagined that would happen eventually. But given that Nitter future is uncertain (wasn't it discontinued a few weeks ago?) and RSS was never the focus, I'm not sure if it's worth investing on this anymore.

You could probably check the current `src/views/rss.nimf` and try to replicate my changes, it's not that much.

thomas-333:
Thanks for the reply gil.

Nitter is working at least for now with real accounts which is fine for my personal use in generating rss feeds. I'm hoping it will continue it at least some form.

After more testing your original code works fine. The problem was with @ninboy changes.

$renderRssTweet(get(tweet.quote), cfg)} should be ${renderRssTweet(get(tweet.quote), cfg)} and the #end if line shouldnot have been deleted. This seems to be working.

PR #825, opened 2023-04-05
added favorites endpoint and added likes tab to profile pages

PrivacyDevel:
Resolves #302

PR #830, opened 2023-04-08
added login-based workaround to view NSFW content

### LINKED ISSUE #302: Feature Request - Likes page
can you implement a likes page

### LINKED ISSUE #829: 18+/"sensitive" tweets no longer available
Any 18+ Tweet that could be viewed before can't be any longer, did some update age restrict them? If so, then how do I switch it off?

PrivacyDevel:
Based on, preferred over and resolves #825
Closes #829 (and closes #302)


OIRNOIR:
If you replace the word Solves with Closes in your initial comment, it will automatically close that issue when this PR is merged.
Since this is based on the changes of another pull request, you may also want to include Closes #302.

roryyamm:
How is this going to work in the end?

PrivacyDevel:
If this pull request got accepted and people updated, they would have the choice to either

- leave everything as is and nothing would change
- create a Twitter account, login and copy the required information from the sent HTTP headers / cookie list of their browser into the nitter.conf file

fucknitter:
Fuck you.

fucknitter:
I think that you should shield people from viewing NSFW content.

fucknitter:
I want to slap you.

OIRNOIR:
> I think that you should shield people from viewing NSFW content.

If you feel it is necessary, you may open an issue requesting the user to confirm their age to access NSFW profiles. I don't know if such a thing will be added anyway.
In general, however, Nitter is not a very well-known project among those who should be protected from viewing NSFW content. The vast majority of those people have never even heard of Nitter.
I would urge parents of underage children to parent responsibly instead of urging passionate developers to weigh down their products with chains.

fucknitter:
I'm very sorry, just having one of those bad days. It doesn't help that my cat is sick and I'm worried regarding his mental state. You must think I'm rude. I'm trying to curb my Twitter addiction and doing this kind of thing to circumvent Elon Musk's poor decisions isn't helping.

src/config.nim
```diff
@@ -40,7 +41,13 @@ proc getConfig*(path: string): (Config, parseCfg.Config) =
     enableRss: cfg.get("Config", "enableRSS", true),
     enableDebug: cfg.get("Config", "enableDebug", false),
     proxy: cfg.get("Config", "proxy", ""),
-    proxyAuth: cfg.get("Config", "proxyAuth", "")
+    proxyAuth: cfg.get("Config", "proxyAuth", ""),
+    cookieHeader: cfg.get("Config", "cookieHeader", ""),
```
bbappserver, line 45: @PrivacyDevel This is inconvenient to use when the values can change around if using it in a docker container.
Please change it to
```nim
cfg.get("Config", "cookieHeader",  getEnv("cookieHeader", "") )
cfg.get("Config", "xCsrfToken",  getEnv("xCsrfToken", "") )
```

bbappserver: When running docker containers through nice web frontends this makes it easy to change the files without needing to screw around with the twitter conf and getting the `cookieHeader` all properly escaped for the `.conf`, because you can just paste it into a text box and it handles the escaping.

somini:
Does this "fix" #919? For public instances this might be hard to implement, but for personal instances, this would be acceptable.

jiatern:
> Does this "fix" #919? For public instances this might be hard to implement, but for personal instances, this would be acceptable.

I just set up a private instance, it works well.

Mendelrock:
> > Does this "fix" #919? For public instances this might be hard to implement, but for personal instances, this would be acceptable.
> 
> I just set up a private instance, it works well.

Would you mind sharing your ways? I tried to set it up too, but so far, I still get route errors. I probably missed a step, but I can't seem to find which :'(

jiatern:
> Would you mind sharing your ways?
I am not sure what you missed, but these are my steps:

1. Clone PrivacyDevel's fork https://github.com/PrivacyDevel/nitter, I use the master branch
2. Install all dependencies as instructed (libpcre, libsass, redis, nimble)
3. Update configurations in nitter.conf, ensure `ct0`, `auth_token` and `xCsrfToken` are also updated
4. Build nitter by following the guidelines
5. Run redis, and then run nitter

I am able to access individual user profiles as well.

fnxweb:
The CSS is _probably_ unrelated to the main functionality not working, but if I fetch http://localhost:8080/css/style.css I actually get an HTML Nitter error not the CSS file.

fnxweb:
CSS fixed, I had neglected the **nimble scss** (which by its name is plainly related!)

Still get 404 out of Twitter though :(

fnxweb:
I started from a git clone of nitter, created a new branch and pulled in your nsfw pull req.  I've not yet tried a direct clone of your branch, but I will.

I copied my working nitter.conf and added the two new  auth fields which I'll partially paste below.  I'll do a more detailed diff of that and your updated example in case I missed something else.  I do have your changes locally, and have rebuilt.  Oh.  But I added some extra debug to print the new fields and that's not coming out, either meaning my config is not being picked up, or it's not somehow rebuilt properly?

# in [Config] section after the comment describing them
cookieHeader = "ct0=e5...ff; auth_token=bb...42"
xCsrfToken = "e5...ff"

That's interesting, the ct0 and Csrf are the same, I'd not noticed that.  Expected?  I'm sure I double checked them.

Version [2023.07.06-bb7e239](https://github.com/zedeus/nitter/commit/bb7e239)

fnxweb:
OK, a raw build of your branch behaves differently, I now get the _error occured in one of your routes_ that @Mendelrock reported above.  And the debug still reports a 404 coming back from the API URL.  I think that error may be a symptom of the 404, as Z's master now has a JSON parser/exception fix that you haven't yet got, I think (but I have in my other build as I merged nsfw onto his latest).

So I still have some issue somewhere.

fnxweb:
Right, not sure what I was doing wrong, but I do get debug now, and upon the 404 I print my auth tokens and they are as I expect.  But I can see zero evidence of them being used.  The only time they are picked up is in the proc fetch (as above) and debug does not show this being called.  I get errors from proc fetchRaw.

It is a bit more complex as well.  If I *search* for the user (_BungieHelp_ FWIW) I do get results.  But if then click on a user name (that or any others), it comes back _not found_ (well, a 404, which Z's master has a fix-ish for, so "not found" instead of the above routes error message).

I'm wondering:  does your patch only add auth to the points where you found it was previously required (like search) and it is not present in the path that is used for a user view / feed (and so that now fails)?

y-bonfire:
I tried this pr today. While RSS retrieval is functioning, it seems that the search functionality is not working properly.
The same situation?
https://github.com/zedeus/nitter/issues/983#issuecomment-1694706348

image

zedeus:
This won't be needed anymore

PR #831, opened 2023-04-09
Escape special characters in long tweets

ftxqxd:
It seems that although the legacy API pre-escapes special characters (`<`, `>`, and `&`), the newer GraphQL API, which is used for long tweets, does not.


zedeus:
Thanks, but this breaks links and hashtag replacements. Fixed in https://github.com/zedeus/nitter/commit/c2819dab441b8ad8220b03dd0fe79f5a5d51b841

PR #837, opened 2023-04-19
raise a RateLimitError when Twitter returns HTTP status 429

PrivacyDevel:


src/apiutils.nim
```diff
@@ -67,6 +67,9 @@ template fetchImpl(result, fetchBody) {.dirty.} =
 
       getContent()
 
+    if resp.status == $Http429:
+      raise rateLimitError()
```
zedeus, line 71:
```suggestion
      if resp.status == $Http429:
        raise rateLimitError()
```

zedeus (changes requested):

PR #885, opened 2023-05-26
fixed bug that caused threads on user profiles to be hidden

PrivacyDevel:


zedeus:
Ah, this is for when you're logged in right? I'm only seeing `homeConversation`, but seems fine to me.

PR #925, opened 2023-07-02
additional bearer tokens & retry faster after rate-limit

drpepper66666:
also need #830, then instance works again


zedeus:
Thank you for your efforts, it was useful in figuring out a good solution. After doing rigorous testing I can conclude that almost all of those bearer tokens are useless, with just 2-3 of them having access to most endpoints.

PR #927, opened 2023-07-10
Fix everything

zedeus:
Uses a new bearer token and a new set of updated endpoints. Only thing missing is search.


GrosMatou27:
and RSS ?

zedeus:
RSS works, it's just disabled on nitter.net

GrosMatou27:
OK, didn't know that. Thank you

Issue #926, opened 2023-07-05
(Probably a failed build) CSS not working?


FurretDev:
![image](https://github.com/zedeus/nitter/assets/99131021/51a07e68-5bd2-46df-aec2-2065539f104b)

The server returns 404 when requesting CSS

zedeus:
From Docker hub?

FurretDev:
> From Docker hub?

nope, i didn't use docker

zedeus:
Weird, maybe a permission thing. I'd need more details to figure it out, along with a way to reproduce.

FurretDev:
> Weird, maybe a permission thing. I'd need more details to figure it out, along with a way to reproduce.

hmm, what exactly do you need? i will try to give you all the necessary information

gfpj:
For me it looked the same, I built on Windows with the same command above, without Docker, redis running in wsl, nim 1.6.12.

FurretDev:
issue solved, i don't really know what i did but it's working now, however...

the search function does not work, and I have noticed that in most of the nitter instances it does not work either, not sure why

zedeus:
I thought "Tweet search is unavailable for now" was obvious enough to make it clear that it isn't broken, it just isn't implemented right now. I'll work on it soon.

osalbahr:
> I thought "Tweet search is unavailable for now" was obvious enough to make it clear that it isn't broken, it just isn't implemented right now. I'll work on it soon.

It was clear to me. However, I am biased since I'm following this issue and already know. I can see how it can be interpreted as "it's broken".

I think the message is good as-is, but something along the lines of "Tweet search isn't implemented right now" would be clearer.

FurretDev:
> I thought "Tweet search is unavailable for now" was obvious enough to make it clear that it isn't broken, it just isn't implemented right now. I'll work on it soon.

i mean it happened on almost every nitter instances so that's why i thought it was broken 😭, but okay

osalbahr:
> i mean it happened on almost every nitter instances so that's why i thought it was broken 😭, but okay

Which instances did nitter search work on?

FurretDev:
> > i mean it happened on almost every nitter instances so that's why i thought it was broken 😭, but okay
> 
> Which instances did nitter search work on?

https://tweet.whateveritworks.org/

this one is **partly functional**, most of the things do not work but for example, i can search users but most of the things will give a 502 error (i think)

PR #935, opened 2023-07-11
Clarify tweet search status

osalbahr:
[#926 ](https://github.com/zedeus/nitter/issues/926#issuecomment-1630832535)


zedeus:
Search has been implemented, so no need for this

PR #952, opened 2023-07-17
Add data-tweet-id attribute to tweet timeline-item and quote elements

floogulinc:
This adds a `data-tweet-id` attribute to the container element of tweets and quotes, including unavailable tweets. This helps when programmatically parsing nitter timelines since there isn't a proper API yet. Especially important is adding the tweet ID to unavailable tweets. I use nitter to grab timelines and then a separate tool to download the individual tweets that supports age-restricted tweets.

### DIFF
--- a/src/views/tweet.nim
+++ b/src/views/tweet.nim
@@ -232,7 +232,7 @@ proc renderQuote(quote: Tweet; prefs: Prefs; path: string): VNode =
         else:
           text "This tweet is unavailable"
 
-  buildHtml(tdiv(class="quote quote-big")):
+  buildHtml(tdiv(class="quote quote-big", "data-tweet-id" = $quote.id)):
     a(class="quote-link", href=getLink(quote))
 
     tdiv(class="tweet-name-row"):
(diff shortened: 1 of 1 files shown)


zedeus:
I've rejected similar PRs that make scraping easier, since excess scraping is a big problem that affects public instances which are not meant to be scraped. A proper API will be added instead.

floogulinc:
That's ironic for a project that scrapes Twitter in any way possible. 

The issue asking for an API and a PR adding one have been open for years so I had no reason to believe it was coming and it's not like an API endpoint would be all that much less intensive than serving the html it currently does, when properly cached. 

If you're concerned about excessive traffic maybe just add some simple rate limits to these routes. That's even something instance owners could do themselves at the reverse proxy level. The "scraping" I would be doing here is just checking nitter timelines every few days or so for new tweets, probably less intensive than many regular users anyway.

zedeus:
It may seem ironic, but Twitter/X is obviously far more capable of handling huge amounts of traffic than public Nitter instances. nitter.net already gets hammered by massive botnets using thousands of IPs at once, and I don't want to make their lives easier. I use nginx rate limits, it's not enough.

Issue #192, opened 2020-06-15
Rest APIs


FireMasterK:
Would be a nice idea to add a Rest API to Nitter.
Invidious's API is a great example of how it could be implemented.

Joshndroid:
I would like to see this also to allow for further front end apps such as fritter on mobile (as I noted wanting to use this as a backend in fritter app for android)

eUgEntOptIc44:
@zedeus @pluja the PR #440 by @xyb seems trying to implement this. For some reason there's no reaction on #440 so far. Would be very grateful for your feeback on this. To me personally nitter is already very useful but the ability of integrations in e.g. mobile apps through a REST api would highly increase the UX. Thank you for maintaining this great project. I'd love to contribute to the code however learning nim upfront is unfortunatley a show stopper as I face resource conflicts.

zedeus:
The PR doesn't cover things like rate limiting, server config to control usage of the API (e.g. disabling it), documentation, reusing code to avoid duplication for every endpoint, etc. I'm not sure yet how to tackle it as it's very hard to change an API after it's made since you end up breaking everything people have built with it, and different Nitter instances would end up having different API versions as well. It needs more consideration and careful effort. I don't have much time for Nitter in general, so this is too much for me to do at the moment.

eUgEntOptIc44:
Ok I see. Sorry to hear that.

r3nor:
Hi, is this feature planned on the roadmap? Would be very helpful when developing on top of Nitter.

PR #978, opened 2023-08-09
feat(api): twitter api router added

kristobalus:


zedeus:
Same issues as the other API PR as described here: https://github.com/zedeus/nitter/issues/192#issuecomment-1214450002
Thank you, but this isn't what Nitter needs.

Issue #1003, opened 2023-08-29
Preferences page — form: label/input for/id relation missing


luce-carevic:
There is no for/id relation between some label and input.

<img width="637" alt="Screenshort of the preferences form" src="https://github.com/zedeus/nitter/assets/6073336/37fda735-ae67-43ce-8bcd-08d84ace00bf">

**Example:**

```
<label for="theme">Theme</label> 
<select name="theme">[…]</select>
```

Should be : 

```
<label for="theme">Theme</label> 
<select name="theme" id="theme">[…]</select>
```

**Similar issue**

In the link replacements section.

<img width="510" alt="Screenshot — code of the link replacements section" src="https://github.com/zedeus/nitter/assets/6073336/95488a22-4872-41f1-bd74-b8194b20d6bc">

Should be : 

- `<input name="replaceTwitter" id="replaceTwitter" type="text" placeholder="Nitter hostname" value="" autofocus="">`
- `<input name="replaceYouTube" id="replaceYouTube" type="text" placeholder="Piped hostname" value="" autofocus="">`
- `<input name="replaceReddit" id="replaceReddit" type="text" placeholder="Teddit hostname" value="" autofocus="">`

Even though associating labels implicitly can be OK, explicit labels are better supported by assistive technology. Consider adding a for/id liaison between label and input.

See: 

- https://www.w3.org/WAI/WCAG21/Techniques/html/H44
- https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-implicitly

Issue #1002, opened 2023-08-29
Liberapay link has no accessible name


luce-carevic:
### Issue 

```
<a href="https://liberapay.com/zedeus"><svg class="lp" viewBox="0 0 40.6 52.3">
  <g transform="matrix(0.83,0,0,0.83,-158,-261)">
    <path d="m202.5,366c-3.1 0-5.5-0.4-7.3-1.2-1.8-0.8-3-1.9-3.8-3.3-0.8-1.4-1.1-3-1.1-4.8 0-1.8 0.3-3.7 0.8-5.8l8.3-34.8 10.2-1.6-9.1 37.8c-0.2 0.8-0.3 1.5-0.3 2.2 0 0.7 0.1 1.2 0.4 1.7 0.3 0.5 0.7 0.9 1.3 1.2 0.6 0.3 1.5 0.5 2.7 0.6l-2 8.1"></path>
    <path d="m239.2 344.3c0 3.2-0.5 6.1-1.6 8.8-1 2.6-2.5 4.9-4.4 6.9-1.9 1.9-4.1 3.4-6.7 4.5-2.6 1.1-5.4 1.6-8.5 1.6-1.5 0-3-0.1-4.5-0.4l-3 11.9h-9.7l10.9-45.4c1.7-0.5 3.7-1 6-1.4 2.3-0.4 4.7-0.6 7.3-0.6 2.4 0 4.6 0.4 6.3 1.1 1.8 0.7 3.2 1.8 4.4 3 1.1 1.3 2 2.8 2.5 4.5 0.5 1.7 0.8 3.6 0.8 5.5m-23.8 13.4c0.7 0.2 1.7 0.3 2.8 0.3 1.7 0 3.3-0.3 4.7-1 1.4-0.6 2.6-1.5 3.6-2.7 1-1.1 1.7-2.5 2.3-4.1 0.5-1.6 0.8-3.4 0.8-5.3 0-1.9-0.4-3.5-1.2-4.8-0.8-1.3-2.3-2-4.3-2-1.4 0-2.7 0.1-3.9 0.4l-4.6 19.1"></path>
  </g>
</svg>
</a>
```

It's an empty link for assistive technologies.

### Solution

```
<a href="https://liberapay.com/zedeus" title="Liberapay — zedeus">
<svg class="lp" viewBox="0 0 40.6 52.3" role="img" aria-label="Liberapay">
  <g transform="matrix(0.83,0,0,0.83,-158,-261)">
[...]

Issue #1001, opened 2023-08-29
About page — <title> should be About | nitter


luce-carevic:
Where : https://nitter.net/about

What should be fixed: `<title>nitter</title>`

How: `<title>About | nitter</title>`

Issue #998, opened 2023-08-29
Main tag missing #a11y


luce-carevic:
People using assistive technologies (like screenreaders) can rely on HTML5 structure tags to navigate more quickly.

There is only one nav tag and no other tags.

Consider transforming the `<div class="container">` into `<main class="container">`.

See: https://www.tpgi.com/easy-content-organisation-with-html5/

PR #1005, opened 2023-08-30
Fix some accessibility issues

hannaeko:
Hello, I am trying to address some of the accessibility issues raised in #998, #1001, #1002 and #1003.

Changes:

* replace div by main for .container element
* change about page title to be "About | nitter.net"
* add role, label and title for liberapay icon / link
* use explicit label to link label to input (this also changed the way checkbox are styled, from a span directly following the checkbox to styling the checkbox directly)
* use separate fieldsets to group inputs in preference form

Cheers,
Gaël


untitaker:
this patch is now deployed on `nitter.woodland.cafe` -- lighthouse a11y score seems to marginally improve on tweets, so I consider that a win

PR #985, opened 2023-08-18
Replace tokens with guest accounts, swap endpoints

zedeus:
This replaces the old method of using guest tokens with using "guest accounts".
For more information see this comment https://github.com/zedeus/nitter/issues/983#issuecomment-1683134811

Live now on nitter.net for testing. If you'd like to test and don't have any guest accounts, feel free to reach out to me on Matrix. A way to acquire guest accounts will be build and shared very soon.

PR #1053, opened 2023-10-31
Bump nim version minimal requirement + update dockerfile arm

unixfox:
From the discussion: https://github.com/zedeus/nitter/pull/985/files#r1377420589

This bumps the minimal requirement for nim version to `1.6.10` and also update the dockerfile for ARM to nim `1.6.14`.


zedeus:
Thanks!

PR #1116, opened 2023-11-14
Add functionality to fetch guest accounts on demand

drpepper66:
Add a basic way to automatically fetch guest accounts from an external URL if the internal stash is empty.

for example:

```
guestAccountsUrl = "https://twitterminator.x86-64-unknown-linux-gnu.zip/download"
guestAccountsKey = "<random string>"
```

If those settings are configured, guest_accounts.json does not need to be present.

this serves the hashed Key under `/.well-known/nitter-auth-sha256`, while twitterminator expects it under `/.twitterminator.txt`. If this approach is accepted I'll update twitterminator to look for the new path too.

### DIFF
--- a/nitter.example.conf
+++ b/nitter.example.conf
@@ -33,6 +33,12 @@ tokenCount = 10
 # always at least `tokenCount` usable tokens. only increase this if you receive
 # major bursts all the time and don't have a rate limiting setup via e.g. nginx
 
+# Instead of guest_accounts.json, fetch guest accounts from an external URL.
+# Nitter will re-fetch accounts if it runs out of valid ones.
+guestAccountsUrl = ""    # https://example.com/download
+guestAccountsHost = ""   # defaults to nitter's hostname
+guestAccountsKey = ""    # a random string that will be used as a secret to authenticate against the URL
+
 # Change default preferences here, see src/prefs_impl.nim for a complete list
 [Preferences]
 theme = "Nitter"
--- a/nitter.example.conf
+++ b/nitter.example.conf
@@ -26,18 +26,14 @@ enableRSS = true  # set this to false to disable RSS feeds
 enableDebug = false  # enable request logs and debug endpoints (/.accounts)
 proxy = ""  # http/https url, SOCKS proxies are not supported
 proxyAuth = ""
-tokenCount = 10
-# minimum amount of usable tokens. tokens are used to authorize API requests,
-# but they expire after ~1 hour, and have a limit of 500 requests per endpoint.
-# the limits reset every 15 minutes, and the pool is filled up so there's
-# always at least `tokenCount` usable tokens. only increase this if you receive
-# major bursts all the time and don't have a rate limiting setup via e.g. nginx
 
 # Instead of guest_accounts.json, fetch guest accounts from an external URL.
 # Nitter will re-fetch accounts if it runs out of valid ones.
-guestAccountsUrl = ""    # https://example.com/download
-guestAccountsHost = ""   # defaults to nitter's hostname
-guestAccountsKey = ""    # a random string that will be used as a secret to authenticate against the URL
+[GuestAccounts]
+usePool = false  # enable fetching accounts from external pool.
+poolUrl = ""     # https://example.com/download
+poolId = ""      # defaults to nitter's hostname
+poolAuth = ""    # random secret string for authentication
 
 # Change default preferences here, see src/prefs_impl.nim for a complete list
 [Preferences]
--- a/src/auth.nim
+++ b/src/auth.nim
@@ -221,14 +221,14 @@ proc updateAccountPool*(cfg: Config) {.async.} =
       let client = newAsyncHttpClient()
 
       try:
-          let resp = await client.get($(cfg.guestAccountsPoolUrl ? {"id": cfg.guestAccountsPoolId, "auth": cfg.guestAccountsPoolAuth}))
-          let guestAccounts = await resp.body
+        let resp = await client.get($(cfg.guestAccountsPoolUrl ? {"id": cfg.guestAccountsPoolId, "auth": cfg.guestAccountsPoolAuth}))
+        let guestAccounts = await resp.body
 
-          log "status code from service: ", resp.status
+        log "status code from service: ", resp.status
 
-          for line in guestAccounts.splitLines:
-            if line != "":
-              accountPool.add parseGuestAccount(line)
+        for line in guestAccounts.splitLines:
+          if line != "":
+            accountPool.add parseGuestAccount(line)
 
       except Exception as e:
         log "failed to fetch from accounts service: ", e.msg
--- a/src/auth.nim
+++ b/src/auth.nim
@@ -218,7 +218,7 @@ proc updateAccountPool*(cfg: Config) {.async.} =
     if accountPool.len == 0:
       log "fetching more accounts from service"
 
-      let client = newAsyncHttpClient()
+      let client = newAsyncHttpClient("nitter-accounts")
 
       try:
         let resp = await client.get($(cfg.guestAccountsPoolUrl ? {"id": cfg.guestAccountsPoolId, "auth": cfg.guestAccountsPoolAuth}))
--- a/src/auth.nim
+++ b/src/auth.nim
@@ -1,5 +1,5 @@
 #SPDX-License-Identifier: AGPL-3.0-only
-import std/[httpclient, asyncdispatch, times, json, random, sequtils, strutils, tables, packedsets, os]
+import std/[httpclient, asyncdispatch, times, json, random, sequtils, strutils, tables, packedsets, os, uri]
 import nimcrypto
 import types, http_pool
 import experimental/parser/guestaccount
@@ -222,7 +222,7 @@ proc updateAccountPool*(cfg: Config) {.async.} =
 
       try:
         pool.use(newHttpHeaders()):
-          let resp = await c.get("$1?id=$2&auth=$3" % [cfg.guestAccountsPoolUrl, cfg.guestAccountsPoolId, cfg.guestAccountsPoolAuth])
+          let resp = await c.get($(cfg.guestAccountsPoolUrl ? {"id": cfg.guestAccountsPoolId, "auth": cfg.guestAccountsPoolAuth}))
           let guestAccounts = await resp.body
 
           log "status code from service: ", resp.status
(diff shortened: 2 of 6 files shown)


nitter.example.conf
```diff
@@ -33,6 +33,12 @@ tokenCount = 10
 # always at least `tokenCount` usable tokens. only increase this if you receive
 # major bursts all the time and don't have a rate limiting setup via e.g. nginx
 
+# Instead of guest_accounts.json, fetch guest accounts from an external URL.
+# Nitter will re-fetch accounts if it runs out of valid ones.
+guestAccountsUrl = ""    # https://example.com/download
+guestAccountsHost = ""   # defaults to nitter's hostname
+guestAccountsKey = ""    # a random string that will be used as a secret to authenticate against the URL
```
zedeus, line 40: Let's put this into a new section, and also remove the `tokenCount` + comment above since that's no longer needed. This section can be used for other variables in the future like whether to try fetching accounts by itself, if so which proxy/proxies to use, jsonl file location, and some other tweaks to the auth handling.

```suggestion
# Instead of guest_accounts.jsonl, fetch guest accounts from an external URL.
# Nitter will re-fetch accounts if it runs out of valid ones.
[GuestAccounts]
usePool = false    # enable fetching accounts from external pool
poolUrl = ""       # https://example.com/download
poolAuth = ""      # random secret string for authentication
poolId = ""        # defaults to nitter's hostname
poolAmount = 10000 # amount of accounts to request from pool
```

src/auth.nim
```diff
@@ -24,6 +25,11 @@ const
   }.toTable
 
 var
+  pool: HttpPool
```
zedeus, line 28: This can be removed, it's really only used to reduce latency for API requests

src/auth.nim
```diff
@@ -24,6 +25,11 @@ const
...
+  guestAccountsUrl = ""
+  guestAccountsHost = ""
+  guestAccountsKey = ""
```
zedeus, line 31: You'll can access these from the cfg object. I suggest setting up a `while true` async loop that tries to get accounts maybe once an hour (tweakable), initialized from the main nitter.nim file. It would be similar to how the token pool used to work: 

https://github.com/zedeus/nitter/blob/b62d73dbd373f08af07c7a79efcd790d3bc1a49c/src/nitter.nim#L41

https://github.com/zedeus/nitter/blob/b62d73dbd373f08af07c7a79efcd790d3bc1a49c/src/tokens.nim#L145-L166

src/auth.nim
```diff
@@ -158,6 +164,27 @@ proc release*(account: GuestAccount) =
   dec account.pending
 
 proc getGuestAccount*(api: Api): Future[GuestAccount] {.async.} =
+  let now = epochTime().int
```
zedeus, line 167: See the comment above for a different way of doing this

src/routes/auth.nim
```diff
@@ -0,0 +1,12 @@
+# SPDX-License-Identifier: AGPL-3.0-only
+
+import jester
+
+import router_utils
+import ".."/[types, auth]
+
+proc createAuthRouter*(cfg: Config) =
+  router auth:
+    get "/.well-known/nitter-auth-sha256":
```
zedeus, line 10: Can we call this something simpler like account-pool-auth?

zedeus (changes requested):

src/routes/auth.nim
drpepper66: i renamed it to nitter-auth. `nitter` because there are some standardized paths under .well-known, and not specific to account-pool since the status page at status.d420.de uses the same auth now

src/auth.nim
drpepper66: done

src/auth.nim
drpepper66: done

src/auth.nim
drpepper66: i moved the http pool into a local variable

nitter.example.conf
drpepper66: done

drpepper66: > poolAmount = 10000 # amount of accounts to request from pool

I did not implement this. twitterminator does not allow you to negotiate the count, not in public api...

src/auth.nim
```diff
@@ -227,20 +198,45 @@ proc initAccountPool*(cfg: Config; path: string) =
...
+
+        log "status code from service: ", resp.status
+
+        for line in guestAccounts.splitLines:
+          if line != "":
+            accountPool.add parseGuestAccount(line)
+
+        accountPool.keepItIf(not it.hasExpired)
+
+    await sleepAsync(3600)
```
zedeus, line 230: This is in milliseconds, I'm assuming you meant `60 * 60 * 1000` for 1 hour

src/auth.nim
```diff
@@ -227,20 +198,45 @@ proc initAccountPool*(cfg: Config; path: string) =
...
   accountPool.keepItIf(not it.hasExpired)
 
   log "Successfully added ", accountPool.len, " valid accounts."
 
+proc updateAccountPool*(cfg: Config) {.async.} =
+  if not cfg.guestAccountsUsePool:
+    return
+
+  while true:
+    if accountPool.len == 0:
```
zedeus, line 214: I think we can improve this a bit using some simple math and some config variables, I'll message you on Matrix.

src/auth.nim
```diff
@@ -227,20 +198,45 @@ proc initAccountPool*(cfg: Config; path: string) =
...
+    await sleepAsync(3600)
+
 proc getAuthHash*(cfg: Config): string =
-  if cfg.guestAccountsKey == "":
-    log "guestAccountsKey is set to bogus value, responding with empty string"
+  if cfg.guestAccountsPoolAuth == "":
+    # If somebody turns on pool auth and provides a dummy key, we should
+    # prevent third parties from using that mis-configured auth and impersonate
+    # this instance
+    log "poolAuth is set to bogus value, responding with empty string"
```
zedeus, line 237: Since this only happens if it's an empty string:
```suggestion
    log "poolAuth is empty, authentication will fail"
```

src/auth.nim
```diff
@@ -227,20 +198,45 @@ proc initAccountPool*(cfg: Config; path: string) =
...
+            accountPool.add parseGuestAccount(line)
+
+        accountPool.keepItIf(not it.hasExpired)
+
+    await sleepAsync(3600)
+
 proc getAuthHash*(cfg: Config): string =
-  if cfg.guestAccountsKey == "":
-    log "guestAccountsKey is set to bogus value, responding with empty string"
+  if cfg.guestAccountsPoolAuth == "":
```
zedeus, line 233:
```suggestion
  if cfg.guestAccountsPoolAuth.len == 0:
```

src/auth.nim
```diff
@@ -227,20 +198,45 @@ proc initAccountPool*(cfg: Config; path: string) =
...
+  if not cfg.guestAccountsUsePool:
+    return
+
+  while true:
+    if accountPool.len == 0:
+      let pool = HttpPool()
+
+      log "fetching more accounts from service"
+      pool.use(newHttpHeaders()):
+        let resp = await c.get("$1?id=$2&auth=$3" % [cfg.guestAccountsPoolUrl, cfg.guestAccountsPoolId, cfg.guestAccountsPoolAuth])
```
zedeus, line 219: If you make `cfg.guestAccountsPoolUrl` a Uri object from std/uri and parse it using `parseUri` in config.nim, you can do:
```suggestion
        let resp = await c.get($(cfg.guestAccountsPoolUrl ? {"id": cfg.guestAccountsPoolId, "auth": cfg.guestAccountsPoolAuth}))
```
It's not a big improvement by any means, but it's more consistent with how URLs are handled in the rest of Nitter's code. The `$(...)` part is only necessary because `safeFetch` from media.nim as mentioned above takes a string instead of a Uri, I'll move and fix the function after this PR gets merged.

src/auth.nim
```diff
@@ -227,20 +198,45 @@ proc initAccountPool*(cfg: Config; path: string) =
...
 
   log "Successfully added ", accountPool.len, " valid accounts."
 
+proc updateAccountPool*(cfg: Config) {.async.} =
+  if not cfg.guestAccountsUsePool:
+    return
+
+  while true:
+    if accountPool.len == 0:
+      let pool = HttpPool()
```
zedeus, line 215: What I meant is that it's quite overkill to use a http pool here, as it keeps open connections sitting around while you don't need them in this case. It should probably be moved to another file, but you can use `safeFetch` from media.nim for now as a simple and easy way to just get the contents without caring about handling the http client or errors.

zedeus (changes requested):

src/auth.nim
drpepper66: i don't think i want that function. I still want to log status code and log exceptions. should I just use the stdlib directly?

zedeus: Sure, that makes sense. Just make sure to do `client.close()` in a `final:` clause.

drpepper66:
some features that are not implemented:

* multi-node setups with load balancers. I think this could be supported by adding a sharding parameter to nitter config (or envvar): two variables ("shard ID", "shard count"), and nitter would only use the guest accounts where `guest_account_id % shard_count = shard_id` -- so you could even use the same guest_accounts.jsonl over multiple nodes

* storing the fetched guest accounts somewhere -- right now nitter will not start up if the accounts service is down, and restarts of nitter will refetch accounts. I was thinking one could just store the accounts in redis to improve reliability and startup speed a bit

src/auth.nim
```diff
@@ -197,13 +198,55 @@ proc initAccountPool*(cfg: Config; path: string) =
...
+
+      try:
+          let resp = await client.get($(cfg.guestAccountsPoolUrl ? {"id": cfg.guestAccountsPoolId, "auth": cfg.guestAccountsPoolAuth}))
+          let guestAccounts = await resp.body
+
+          log "status code from service: ", resp.status
+
+          for line in guestAccounts.splitLines:
+            if line != "":
+              accountPool.add parseGuestAccount(line)
```
zedeus, line 231: Minor change here for consistency + indent fix:
```suggestion
      let client = newAsyncHttpClient()

      try:
        let 
          resp = await client.get(cfg.guestAccountsPoolUrl ? {"id": cfg.guestAccountsPoolId, "auth": cfg.guestAccountsPoolAuth})
          guestAccounts = await resp.body

        log "status code from service: ", resp.status

        for line in guestAccounts.splitLines:
          if line.len == 0:
            accountPool.add parseGuestAccount(line)
```

drpepper66: oops!

zedeus: Also, you can add a user-agent if you want, just `newAsyncHttpClient("...")`

Issue #1147, opened 2024-01-10
Videos not playing on nitter.net


WOnder93:
They were still working a couple days ago. Any idea what's wrong?

ElTioRata:
Some videos that are a few seconds long do play for me but longer videos don't.

tonytonyistony:
Experiencing the same issue for the past few days

WOnder93:
Actually I found [a (long) video](https://nitter.net/rabbit_hmi/status/1744781083831574824#m) that seems to play fine. OTOH many others like [this one](https://nitter.net/ResoluteSquare/status/1744752186121163058#m) still won't play.

rcsavo:
In my case, deactivating the proxy video streaming and a hard refresh (Ctrl + Shift + r) got the videos working again.

tonytonyistony:
> In my case, deactivating the proxy video streaming and a hard refresh (Ctrl + Shift + r) got the videos working again.

Confirmed. This worked for me.

ElTioRata:
Yep. Seems like issue is with the proxy. Works flawlessly after disabling that option.

jackyzy823:
Two bugs here.

1. nitter currently not support proxying `cmaf` container : because 
    `#EXT-X-MEDIA:NAME="Audio",TYPE=AUDIO,GROUP-ID="audio-128000",AUTOSELECT=YES,URI="/ext_tw_video/xxxxx/pu/pl/mp4a/128000/xxxxxx.m3u8?container=cmaf"` is not proxied in [`proxifyVideo`](https://github.com/zedeus/nitter/blob/b62d73dbd373f08af07c7a79efcd790d3bc1a49c/src/formatters.nim#L80-L89).
     
3.  current hls.js version might not work well with this container.  (the latest 1.5.1 works well)

zedeus:
Fixed in https://github.com/zedeus/nitter/commit/52db03b73ad5f83f67c83ab197ae3b20a2523d39, live on nitter.net

PR #1148, opened 2024-01-11
Fix for #1147, Proxy for audio URL and upgrade hls.js

jackyzy823:
Fix for #1147, 
1. Proxy for audio URL in `cmaf` container
2. Upgrade `hls.js` to 1.5.1 and use full version

-----
Or , we could force using `container=fmp4` instead of c`ontainer=cmaf`


zedeus:
Thank you for the quick fix, but I'd like to avoid the larger hls.js file and stick with the previous behavior. I've pushed a simple fix to force fmp4.

jackyzy823:
It's ok for me.  Just afraid that twitter might drop `fmp4` support in the feature.

jackyzy823:
> Just afraid that twitter might drop `fmp4` support in the feature.

Sadly, It happened. event using fmp4 in the query , it still returns m3u8 with cmaf format.

But whatever.

zedeus:
Getting complaints about broken videos, and this supposedly fixes it. Can you reopen and rebase please?

PR #1185, opened 2024-02-26
declare dead in README

qpwo:


ecxod:
It is not dead at all.

PR #1298, opened 2025-10-19
Enable using of configured http proxy for media

digletthat:
Currently, configured http proxy (`Config.proxy`, `Config.proxy_auth`) is only used for API calls. Media (`twimg.com`) is always downloaded directly. This PR adds usage of proxy for media too.

This PR adds some code duplication: both `http_pool` and `routes/media` modules now have `setHttpProxy` method and `proxy` var. Maybe I should refactor this a bit and make single global var and setter somewhere, but I'm not sure where  to put it.

Also, maybe there should be separate proxy settings for API and media in config?

This is my first code edits in nim language.


digletthat:
> Does this allow to toggle on or off in the preferences?

No. I don't think it should be configurable by user. It's related to http proxy that is used for outgoing http requests, so it's part of nitter server configuration.

unixfox:
Now that I think again about this PR, I didn't fully understand it well.

It's not about activating or deactivating the media proxy in nitter. But the http proxy when used in nitter. Sorry.

PR #1305, opened 2025-11-16
fix: correct argument parsing for --append flag in get_web_session.py

0xcathiefish:
## Description
Fixes argument parsing bug in `tools/get_web_session.py` when using `--append` flag together with TOTP seed parameter.

## Problem
The current implementation uses `enumerate(sys.argv[3:], 3)` which causes incorrect index calculation when accessing `sys.argv[i + 1]` for the `--append` parameter value. This results in the `--append` flag not working properly when a TOTP seed is provided.

**Example that fails with current code:**
```bash
python3 tools/get_web_session.py myuser mypass "TOTP_SECRET" --append sessions.jsonl
```

## Root Cause
When using `enumerate(sys.argv[3:], 3)`, the index `i` refers to the position in the original `sys.argv`, but when trying to access `sys.argv[i + 1]`, the calculation becomes incorrect due to the offset introduced by slicing.

## Changes Made
- Replaced `enumerate()` loop with explicit `while` loop and manual index increment
- Added `if totp_seed is None` check to prevent overwriting TOTP seed with subsequent non-flag arguments
- Added error handling when `--append` is provided without a filename
- Added missing `import os` for the `os._exit()` call
- Improved argument parsing to correctly skip consumed arguments

## Testing
Tested with the following commands to ensure correct behavior:
```bash
# With TOTP and append
python3 tools/get_web_session.py user pass "TOTP_SECRET" --append sessions.jsonl

# With all flags
python3 tools/get_web_session.py user pass "TOTP_SECRET" --headless --append sessions.jsonl

# Without TOTP
python3 tools/get_web_session.py user pass --append sessions.jsonl
```

All argument combinations now parse correctly.

### DIFF
--- a/tools/get_web_session.py
+++ b/tools/get_web_session.py
@@ -77,9 +77,9 @@ async def login_and_get_cookies(username, password, totp_seed=None, headless=Fal
                     twid = cookies_dict['twid']
                     # Try to extract the ID from twid (format: u%3D<id> or u=<id>)
                     if 'u%3D' in twid:
-                        user_id = twid.split('u%3D')[1].split('&')[0]
+                        user_id = twid.split('u%3D')[1].split('&')[0].strip('"')
                     elif 'u=' in twid:
-                        user_id = twid.split('u=')[1].split('&')[0]
+                        user_id = twid.split('u=')[1].split('&')[0].strip('"')
 
                 cookies_dict['username'] = username
                 if user_id:


0xcathiefish:
## Additional Fix

I discovered another bug while testing: the extracted `user_id` from the `twid` cookie includes trailing quotes, resulting in invalid JSON output.

**Example of buggy output:**
```json
{"kind": "cookie", "username": "...", "id": "111222333\"", ... }
```

**Fix applied:**
Added `.strip('"')` to remove quotes when extracting user_id:
```python
if 'u%3D' in twid:
    user_id = twid.split('u%3D')[1].split('&')[0].strip('"')
elif 'u=' in twid:
    user_id = twid.split('u=')[1].split('&')[0].strip('"')
```

**Verified output:**
```json
{"kind": "cookie", "username": "...", "id": "111222333", ... }
```

The script now works correctly and generates valid JSON. ✅
This has been added to the same PR in commit [commit_hash](https://github.com/zedeus/nitter/pull/1305/commits/3190623cff2c9855743d20d5fa6dc38e140be384).

PR #1347, opened 2025-12-23
Fix the checkmark position

796176:
## Description

If the profile name is too long and has a checkmark, this checkmark gets in the way of other elements on the right. On mobile it's even worse because it widens the page by creating an empty space on the right side to fit the checkmark.

## Reproduction

Open any page with a tweet from a verified author with a long profile name like [this](https://nitter.net/mushoku_eiyu/status/2001283991341871371) or [this](https://nitter.net/futsutsuka_PR/status/2003299652867867082).

## Solution

The fix is trivial and requires moving the verified-icon div element one level up.

I built and run the patched version, which took me way more time than i expected, to make sure everything works correctly. The upper screenshot is taken using nitter.com and the lower is taken using the patched version.

### DIFF
--- a/src/views/profile.nim
+++ b/src/views/profile.nim
@@ -26,6 +26,7 @@ proc renderUserCard*(user: User; prefs: Prefs): VNode =
 
       tdiv(class="profile-card-tabs-name"):
         linkUser(user, class="profile-card-fullname")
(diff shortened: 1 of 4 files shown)


zedeus:
excellent, thank you!

PR #1400, opened 2026-06-01
Fix OpenSearch response crash

zestysoft:
## Summary

Fixes a crash when serving `/opensearch` by binding the response headers before passing them to `resp`.

The route previously passed an inline header literal directly to `resp`:

```nim
resp Http200, {"Content-Type": "application/opensearchdescription+xml"}, ...
```

With Nim 2.2.0/Jester, requesting `/opensearch` could SIGSEGV while handling that response. Binding the headers to a local first avoids that assignment path while preserving the same response status, content type, and XML body.

## Validation

- Reproduced the crash with a direct request to `/opensearch`.
- A debug build showed the crash at `src/routes/search.nim` inside the `/opensearch` response path, through Nim's generic assignment handling.
- Verified the patched route returns `200 OK` with `application/opensearchdescription+xml` and the expected XML.

### DIFF
--- a/src/routes/search.nim
+++ b/src/routes/search.nim
@@ -46,6 +46,7 @@ proc createSearchRouter*(cfg: Config) =
       redirect("/search?f=tweets&q=" & encodeUrl("#" & @"hash"))
 
     get "/opensearch":
-      let url = getUrlPrefix(cfg) & "/search?f=tweets&q="
-      resp Http200, {"Content-Type": "application/opensearchdescription+xml"},
-                     generateOpenSearchXML(cfg.title, cfg.hostname, url)
+      let
+        url = getUrlPrefix(cfg) & "/search?f=tweets&q="
+        headers = {"Content-Type": "application/opensearchdescription+xml"}
+      resp Http200, headers, generateOpenSearchXML(cfg.title, cfg.hostname, url)


zestysoft:
Functionally, the HTTP response is intended to be identical. The difference is evaluation/lifetime, not business logic.

  Before:

  resp Http200, {"Content-Type": "application/opensearchdescription+xml"},
                 generateOpenSearchXML(...)

  That creates the header container inline as a temporary expression and passes it directly into Jester’s resp macro/template inside an async route. In the failing build, Nim/Jester trips while assigning/copying that temporary header value. The debug stack showed the crash going through Nim’s generic assignment code from this line.

  After:

  let
    headers = {"Content-Type": "application/opensearchdescription+xml"}

  resp Http200, headers, generateOpenSearchXML(...)

  This gives the header container a named local variable before resp sees it. So Jester receives a normal local value instead of an inline aggregate temporary. That avoids the bad assignment/copy path that was causing the SIGSEGV.

  So the user-visible behavior is unchanged:

  HTTP 200
  Content-Type: application/opensearchdescription+xml
  same XML body

  The fix is about making the value passed to resp stable and explicit, avoiding a compiler/runtime/Jester edge case with inline header literals in async route code.

zedeus:
Oh wow, great catch. Thank you so much!

PR #1412, opened 2026-06-17
Exclude git from docker context

fa0311:
I couldn’t build the Docker image in my environment without this.


tiekoetter:
This kills the instance version info display. Just FYI.

zedeus:
Oh, that's not great

fa0311:
After cloning again, this issue no longer reproduced.
It is likely that this is a Nimble bug.
may need to revert it.

PR #1432, opened 2026-08-04
fix: update browser session login flow

NegiKirin:
## Summary
- locate Chrome or Chromium explicitly before starting Nodriver
- update the X username selector to `input[name="username_or_email"]`
- click the current login form’s Continue button after each credential step
- wait for an interactive password field instead of using the hidden initial field

## Validation
- `.venv/bin/python -m py_compile tools/create_session_browser.py`
- live X login form check with Playwright: username selector, visible enabled Continue button, and password field exist
- Nodriver smoke test with a non-existent username confirmed the Continue button is clicked and X returns its expected unknown-account response

No credentials were used in validation.

### DIFF
--- a/tools/create_session_browser.py
+++ b/tools/create_session_browser.py
@@ -23,6 +23,7 @@ Output:
 import asyncio
 import json
 import os
+import shutil
 import sys
 
 import nodriver as uc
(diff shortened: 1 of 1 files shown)


zedeus:
Thank you, but this wasn't quite enough to adapt to the new login flow. Replaced by https://github.com/zedeus/nitter/commit/607765c463e78b80e85ebe797534a5e56affdf94

